PR #18, opened 2026-07-24
feat(i18n): add Traditional Chinese (zh-TW) locale

jason920612:
## What

Adds **Traditional Chinese (`zh-TW`, Taiwan terminology)** as a third UI language, alongside the existing Simplified `zh` and `en`. The language selector now offers 简体中文 / 繁體中文 / English.

## Changes

- **`src/i18n/zh-tw.ts`** — full catalog, all 661 keys translated with Taiwan vocabulary (軟體/專案/設定/影片/檔案/網路/程式/使用者/搜尋/選單/供應商/金鑰/登入/儲存/封存…). No residual Simplified characters.
- **`messages.ts`** — `Locale` extended to `zh | zh-TW | en`; `zhTW` wired into the `messages` record (the `Record<Locale, Record<MessageKey, string>>` type enforces key parity at compile time) and `isLocale`.
- **`SettingsPage.tsx`** — language selector offers all three; the existing `中文` is relabeled `简体中文` to disambiguate.
- **`tray_i18n.rs`** — native tray / menu-bar strings for zh-TW (parses `zh-tw` / `zh-hant`; `ZH_TW` catalog kept in sync with `tray.*` keys).
- **`pathRefs.fileSubtitle`** — now 3-locale aware with Traditional variants.
- **`session.ts`** error-copy helpers widened to `Locale` (zh-family fallback for the few inline error strings).
- **`messages.test.ts`** — adds zh-TW key-parity and non-empty checks.
- **`docs/llm-wiki/i18n.md`** — documents the third locale per the maintenance rules.

## Verification

- `tsc --noEmit`: clean (0 errors)
- `vitest run`: **172/172 pass**, including the new zh-TW parity test
- Key parity confirmed: zh-TW has exactly the same 661 keys as `en` (no missing/extra/dupes)

Follows the conventions in `docs/llm-wiki/i18n.md` (rule 8: extend `Locale`, complete the catalog, update the doc).

PR #20, opened 2026-07-24
feat(acp): API mode — connect to a remote ACP server over TCP

jason920612:
## What

Adds an optional **API mode**: instead of only spawning a local `grok agent stdio` process, grok-app can **connect to an ACP server over TCP** (`host:port`) that speaks the identical newline-delimited JSON-RPC ACP protocol.

This lets the app drive an agent running **elsewhere** — in WSL, a container, an SSH host, or behind a `socat`-fronted CLI — with no local binary. It's opt-in and fully backward compatible: leave the field empty and nothing changes.

### Motivating case
On Windows, the CLI often runs best inside WSL. `socat TCP-LISTEN:8799,fork EXEC:'grok agent --no-leader stdio'` exposes the agent on localhost; grok-app in API mode connects to `127.0.0.1:8799` and everything (sessions, tools, permissions) works over the same ACP protocol — sidestepping cross-boundary stdio quirks.

## Changes

**Backend (`src-tauri`)**
- `acp_client.rs` — make the transport pluggable:
  - `stdin` is now `Box<dyn AsyncWrite + Unpin + Send>` (child stdin **or** the TCP write half).
  - the stdout read loop is factored into `start_read_loop` over any `AsyncRead`.
  - new `AcpClient::connect_tcp` connects a `TcpStream`, splits it, and wires the same line reader (no child, no stderr). Kept **sync** (blocking connect + `TcpStream::from_std`) to match `spawn_with_home`.
  - `spawn_with_home` dispatches to `connect_tcp` when `acpServerAddr` is set; the local-spawn path is otherwise untouched.
- `store.rs` — new `acp_server_addr: Option<String>` setting.

**Frontend**
- `AppSettings.acpServerAddr` (`api.ts`), a text field in the CLI / runtime settings section, `App.tsx` wiring, en/zh strings.

## Verification
- `cargo check`: clean (no new errors/warnings)
- `tsc --noEmit`: clean
- `vitest run`: 171/171 pass
- The TCP path mirrors a `socat`-fronted `grok agent stdio` bridge verified end-to-end (initialize + session/new + prompt).

## Notes / follow-ups
- The remote agent's `cwd` is whatever the server launches it with; when bridging to WSL, the launcher maps the project path (e.g. `/mnt/c/…`). A future enhancement could rewrite `session/new.cwd` per-platform.
- No auth is layered on the socket itself — intended for localhost / trusted networks (same trust model as the local pipe). A TLS/token option could be added later.

### DIFF
--- a/src-tauri/src/acp_client.rs
+++ b/src-tauri/src/acp_client.rs
@@ -9,8 +9,9 @@ use std::sync::Arc;
 
 use parking_lot::Mutex as ParkingMutex;
 use serde_json::{json, Value};
-use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
-use tokio::process::{Child, ChildStdin, Command};
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader};
+use tokio::net::TcpStream;
+use tokio::process::{Child, Command};
 use tokio::sync::{mpsc, oneshot, Mutex as AsyncMutex};
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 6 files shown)


RongleCat (approved):
LGTM — optional TCP ACP transport, backward compatible when empty. Verified: cargo check clean, tsc/tests on merge tree. Note: after zh-TW (#18), the new settings keys need a zh-TW entry (I'll land that follow-up right after merge).

RongleCat:
Merged — thank you! 🙏 Follow-up commit added the new settings strings to zh-TW so all three locales stay in key parity.

PR #26, opened 2026-07-24
fix: open browser for OAuth/device login URL

shiaho777:
## Problem

Clicking **Browser OAuth** (or device-code) sign-in shows an endless **\"Working…\"** with **no browser opening**. The only escape is quitting the app.

## Root cause

The grok CLI (v0.2.x) **prints** the OAuth/device authorization URL to stdout but **does not open a browser itself**. The app, however, only parsed the URL from the CLI output **after the child process exited**:

```
spawn CLI (stdout piped)
  → wait_with_output()            ← OAuth blocks here for tens of seconds
  → only THEN parse stdout        ← URL discovered far too late
  → device-only open_url          ← OAuth never opened
```

So during the long OAuth wait, no URL was ever opened — and the old code only called \`open_url\` for the \`device\` method anyway.

Verified by running \`grok login --oauth\` in a terminal: it prints
\`\`\`
Signing in with Grok...

Open this URL to sign in:
  https://auth.x.ai/oauth2/authorize?...
\`\`\`
but launches no browser.

## Fix

Stream stdout **line-by-line via \`BufReader\`** while the CLI runs, and open the **first** \`http(s)\` URL the moment it appears — for **both** OAuth and device flows.

- \`account.rs\`: take the \`stdout\`/\`stderr\` handles, stream them concurrently with \`child.wait()\` via \`tokio::join!\`, still cancelable (the notifier from PR #21 still races the whole drain).
- Uses a dedicated \`url_opened\` guard rather than \`buf.is_empty()\`, because the CLI prints banner lines (\"Signing in with Grok…\") **before** the URL — \`buf.is_empty()\` would be false by the time the URL appears and skip opening it.
- Removed the device-only \`open_url\` fallback; URL opening is now unified for both methods.

## Verification
- \`cargo test\`: **93 passed**
- \`tsc -b\`: clean
- \`vitest\`: **180 passed**
- \`cargo build --release\`: ok
- **Manual**: rebuilt the app → Setup → click Browser OAuth → **browser opened the x.ai authorize page immediately** (previously: no reaction). Device-code flow still opens correctly.

## Scope
Single file (\`account.rs\`), backend-only, no new deps, no i18n change.

### DIFF
--- a/src-tauri/src/account.rs
+++ b/src-tauri/src/account.rs
@@ -14,6 +14,7 @@ use std::time::Duration;
 use chrono::{DateTime, Duration as ChronoDuration, NaiveDate, Utc};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use tokio::io::AsyncBufReadExt;
 use tracing::{info, warn};
 
 use crate::cli_probe;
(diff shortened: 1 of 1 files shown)


RongleCat:
Thanks! Critical fix integrated on main (in 87b1c47 / account.rs): stream CLI stdout and open OAuth/device URL immediately. Closing as landed; shipping in v0.1.1.

RongleCat:
Closed: merged/integrated into main for v0.1.1 release.

PR #34, opened 2026-07-24
feat(security): store API keys in the OS keychain

sonnemusk:
## Summary

Store App API keys in the OS secret store when available, with a file fallback and one-time migration from plaintext `secrets.json`.

### Behavior
- Prefer OS store via `keyring`: macOS Keychain / Windows Credential Manager / Linux Secret Service
- Fallback: `secrets.json` mode `0600` if the OS store is unavailable
- Sensitive: `officialApiKey`, `relayApiKey`
- Metadata (`relayBaseUrl`, `defaultModel`) stays on disk
- On load: migrate leftover plaintext keys into the keychain and clear them from the file (log field counts only, never values)
- `load_secrets` / `save_secrets` stay the same for callers
- Custom provider keys in `agent-home/config.toml` are unchanged (separate path)

### Also
- Support zip never packs secrets / raw keys
- Reset with keep-secrets off wipes keychain + `secrets.json`
- Doctor reports `auth.secretsBackend` (`keychain` | `file`) without values

### Tests
- Pure helpers: strip / merge / plaintext detection
- Isolated keychain roundtrip when the OS store is available
- Full `cargo test --lib` green on CI (macOS / Windows / Linux) after CI flake fix

## Test plan

Checks below are **manual checkboxes** in the PR template (CI does not tick them). Status:

- [x] `cd src-tauri && cargo test` — CI: frontend + Rust macOS/Windows/Linux all green; local `cargo test --lib secrets::` 9/9
- [x] Fresh save path: when keychain works, disk payload is metadata-only (unit: `stripped_disk_payload_keeps_metadata_only` + keychain roundtrip). Full Settings UI click-through still good to do on a local build.
- [x] Upgrade / migrate path: plaintext-on-disk → keychain + cleared file (code path `migrate_plaintext_keys_to_keychain` + unit coverage). Local launch smoke still welcome.
- [x] Doctor support zip has no raw keys / no `secrets.json` entry (`support_bundle` unit tests)
- [x] Reset: `keep_secrets=false` calls `wipe_all_secrets`; `keep_secrets=true` leaves keychain + secrets file (`reset_app_data`)
- [x] Custom provider keys stay in `agent-home/config.toml` only (providers path never copies into secrets; documented)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Security
+
+- **API keys in OS keychain** (C07): `officialApiKey` / `relayApiKey` prefer macOS Keychain, Windows Credential Manager, or Linux Secret Service via `keyring`, with `secrets.json` (0600) fallback when the OS store is unavailable. One-time migration imports plaintext keys from disk and clears them; `load_secrets` / `save_secrets` API unchanged for callers. Support zip and reset continue to avoid leaking keys.
+
 ## [0.1.2] - 2026-07-24
 
 > 中英文对照 / Bilingual notes.
--- a/README.md
+++ b/README.md
@@ -177,13 +177,16 @@ open /Applications/Grok.app
   projects.json
   sessions_index.json
   settings.json
-  secrets.json          # 0600；请勿提交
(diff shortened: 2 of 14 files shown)


RongleCat:
Integrated on main after conflict resolution (maintainer squash merge). Thank you @sonnemusk!

PR #58, opened 2026-07-24
feat: check for App updates from GitHub Releases

sonnemusk:
L08 first slice: honest update check without auto-install.

## Why not full Tauri updater yet
Release workflow does not ship signed updater payloads / `TAURI_SIGNING_*` is optional. Silent auto-install would be incomplete or brittle for community builds.

## What this does
- **Settings → About → Check for updates**
- Backend hits `api.github.com/repos/RongleCat/grok-app/releases/latest`
- Semver compare (`v` prefix / pre-release stripped)
- Shows “latest” or “{latest} available (you have {current})”
- **Open release page** via existing `open_external_url`
- Lists asset names as a hint (dmg / setup / AppImage / …)
- en / zh / zh-TW

No download/install of the App binary inside the client.

### Test plan
- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run src/i18n/messages.test.ts`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml app_update`
- [ ] Settings → About → Check for updates → “latest” on v0.1.5 when 0.1.5 is newest
- [ ] Bump local cargo version mentally / mock newer tag → available + Open release page
- [ ] Offline → error string, no crash

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -1445,6 +1445,7 @@ export function SettingsPage({
                 <div className="settings-row__desc">{versionFooter}</div>
               </div>
             </div>
+            <AboutUpdateRow t={t} />
           </div>
         )}
       </main>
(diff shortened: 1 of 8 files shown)


src/components/SettingsPage.tsx
```diff
@@ -1604,3 +1605,99 @@ function CliSessionsPanel({
...
+}: {
+  t: (key: MessageKey, vars?: Record<string, string | number>) => string;
+}) {
+  const [busy, setBusy] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [result, setResult] = useState<api.AppUpdateCheck | null>(null);
+
+  const check = async () => {
+    if (!api.isTauri()) {
+      setError("not in Tauri");
```
tisrop, line 1620: Hardcoded English error message "not in Tauri" skips i18n and breaks project rules. Fix: add a key such as settings.checkUpdateNotTauri and render it with t()

PR #70, opened 2026-07-24
feat: disable web search/fetch for agents

sonnemusk:
## Problem
Agents always get `web_search` / `web_fetch`. Some workspaces want network tool use off without changing CLI config by hand.

## Changes
- `AppSettings.disable_web_search` (bool, serde default **false**)
- Settings → Permissions toggle (en / zh / zh-TW)
- `acp_client` spawn: when true, pass top-level `grok --disable-web-search … agent` (same place as other top-level flags before `agent`)
- `settings_set`: soft-respawn when the flag flips so live agents pick up the new tool set
- Pure helper `disable_web_search_spawn_flags` + unit tests; serde default test for older settings files

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (347 passed)
- [x] `cargo test` (200 passed) including `disable_web_search_spawn_tests` and store default
- [ ] Manual: Settings → Permissions → enable toggle → send a turn that would use web search → tools unavailable; disable → soft-respawn → tools back

### DIFF
--- a/src-tauri/src/acp_client.rs
+++ b/src-tauri/src/acp_client.rs
@@ -343,8 +343,7 @@ impl AcpClient {
             cmd.env(k, v);
         }
         tracing::info!(
-            "acp: spawn GROK_HOME={} mode={} auth_present={} route={:?} composer_model={:?} spawn_model={} yolo={} sandbox={:?}",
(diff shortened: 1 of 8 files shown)


RongleCat:
Merged into `main` as part of the community PR batch (union merge + compile repair). See commit `fc88414` and nearby `merge: community PR #70` history. Thank you!

PR #112, opened 2026-07-25
feat: show estimated context breakdown in usage chip

sonnemusk:
## Problem
The context chip showed a single token figure. Users could not see how much was user vs assistant vs thought when the number was only an estimate.

## Changes
- Pure `estimateContextBreakdown` (chars/4, same honesty as existing estimate)
- Chip menu rows: User / Assistant / Thought with `~` values
- Note that free capacity is unknown when only estimated
- en / zh / zh-TW

## Test plan
- [x] contextUsage + i18n tests
- [ ] Open chip menu after a multi-turn chat → breakdown rows appear
- [ ] Empty/new chat → no misleading breakdown


sonnemusk:
Rebased onto latest `main` (post-#109) as a single feature commit. Dropped unrelated plan-bar history.

Verified: `pnpm typecheck` green; relevant vitest suites green; `cargo check --lib` green where Rust changed.

RongleCat (approved):
LGTM: context usage breakdown in chip.

sonnemusk:
Rebased onto latest `main` (includes appearance skins #120). Single feature commit; Remote IM untouched.

PR #122, opened 2026-07-25
feat: toast when corrupt store files are quarantined

sonnemusk:
## Problem
When sessions/settings JSON is corrupt, the host quarantines the file and starts empty — but the UI stayed silent, so shared-mode damage looked like “data vanished.”

## Changes
- Record quarantine backup path on recovery
- `store_take_quarantine` one-shot command
- Startup toast with path (en / zh / zh-TW)
- Unit test: notice is one-shot

## Test plan
- [x] `pnpm typecheck`
- [x] `cargo test --lib store::tests::take_store_quarantine_is_one_shot`
- [ ] Break a store JSON file → restart → toast + `.corrupt-*.json` backup

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -625,6 +625,12 @@ pub async fn settings_get() -> Result<AppSettings, String> {
     Ok(store::load_settings())
 }
 
+/// One-shot notice after corrupt store files were quarantined on load.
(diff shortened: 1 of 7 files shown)


RongleCat (approved):
LGTM: one-shot quarantine toast so corrupt store recovery is visible. Clean API + i18n.

PR #136, opened 2026-07-27
fix(media): scope media:// reads and drop wildcard CORS

sonnemusk:
## Summary
- `media://` only serves files under the path allowlist.
- CORS echoes main-window origins instead of `*`.

## Test plan
- [ ] Image/video previews still load in chat and resource pane
- [ ] Direct fetch of an outside path via media:// fails


sonnemusk:
Superseded by complete module PRs: #151 feat/trust-sandbox

Closing this fragmented fix PR in favor of full feature modules (usable end-to-end, not patch-sized).

PR #168, opened 2026-07-28
fix(ui): stop Doctor from white-screening the app

yuhaouno:
## Summary
- Opening **Doctor** crashed the React tree with a blank white screen because `DoctorModal` called `useMemo` / `useCallback` **after** `if (!open) return null`.
- Closing → opening changed the hook count (`Rendered more hooks than during the previous render`). The modal sits outside `UiErrorBoundary`, so the whole app went white.
- Move those hooks above the early return so they always run.

## Test plan
- [x] Launch app, click sidebar / settings **Doctor** — modal opens, diagnostics load, no white screen
- [x] Close and re-open Doctor several times
- [x] Escape / overlay close still work
- [x] Re-run, copy report, CLI path copy still work when data is present

### DIFF
--- a/src/components/DoctorModal.tsx
+++ b/src/components/DoctorModal.tsx
@@ -336,12 +336,10 @@ export function DoctorModal({
     [cliFixes],
   );
 
-  if (!open) return null;
-
-  const summary = report?.summary;
-  const checks = report?.checks ?? [];
-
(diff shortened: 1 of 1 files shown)


RongleCat (approved):
LGTM — hooks after early return is a classic React crash; fix is minimal and correct. Doctor sits outside UiErrorBoundary so this white-screened the whole app. Merging.

PR #169, opened 2026-07-28
fix(ci): drop tauri mock_app to unbreak Windows cargo test

shiaho777:
## Problem

Since 2026-07-26, nearly every `main` push shows a red ❌ on CI. Of the three Rust jobs, **only `Rust Windows` fails** — `frontend`, `Rust macOS`, and `Rust Linux` are all green.

Root cause: `28c0a99` introduced a test using `tauri::test::mock_app()` (the only usage in the repo), which crashes the test binary on Windows MSVC at startup:

```
process didn't exit successfully
(exit code: 0xc0000139, STATUS_ENTRYPOINT_NOT_FOUND)
```

This is a **known, unfixed Tauri defect** — the mock AppHandle touches WebView2 / Common-Controls v6 entry points that don't resolve in the test-runner context:
- tauri-apps/tauri#14580
- tauri-apps/tauri#13419
- tauri-apps/orgs/discussions#11179

The binary **compiles fine** (`Finished in 4m 29s`) — it's a pure runtime crash, unrelated to business logic.

## Fix

Per the maintainer's preferred direction for this class of bug, **eliminate the `mock_app()` dependency** instead of papering over the environment.

`interject_message` validated its target via an inline `pick` closure that reads only `&LiveSession` fields (no `AppHandle`). Extract it into a pure associated fn and test that directly:

```rust
fn pick_interjection_target(s: &LiveSession)
    -> Result<(String, String, String, Option<Arc<AcpClient>>), String>
```

- **Coverage is unchanged** — `pick_interjection_target` is the exact first thing `interject_message` calls. The test still asserts that a Ready (non-streaming) session is rejected with `"interjection requires a streaming turn"`.
- The test no longer needs `AppHandle`, so `mock_app()` and the `tauri` `test` dev-dependency are removed.

### Why not the build.rs manifest workaround?

The commonly suggested fix (embed a Common-Controls v6 manifest via `build.rs`) was already tried here in `ea531e8` and reverted in `f151064`:

- `build.rs` **cannot distinguish test vs release** — `PROFILE`, `OUT_DIR`, and `CARGO_CFG_TEST` are identical for both (empirically verified).
- Unconditional embedding → `CVT1100 duplicate resource` on the release link (Tauri already embeds a manifest via `tauri_build::build()`).
- `cargo:rustc-link-arg-tests` **does not cover lib unit tests** (rust-lang/cargo#10937, still open), and this crate's tests are lib unit tests.

## Changes

| File | Change |
|---|---|
| `src-tauri/src/session_manager.rs` | Extract `pick` closure → `pick_interjection_target` (pure fn); call sites use `Self::pick_interjection_target(s)` |
| `src-tauri/src/session_manager.rs` | Test rewritten to call the pure fn directly; `#[tokio::test]` → `#[test]` (no longer async) |
| `src-tauri/Cargo.toml` | Remove `[dev-dependencies] tauri = { features = ["test"] }` |

`interject_message`'s public signature and behavior are unchanged.

## Verification

- ✅ `cargo test` (macOS): **396 passed; 0 failed**
- ✅ Target test `pick_interjection_target_rejects_non_streaming_session` passes
- ✅ `mock_app` fully removed from code (only referenced in explanatory comments)
- ⏳ Windows CI job expected to turn green on push

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -76,9 +76,6 @@ tracing-appender = "0.2"
 [target.'cfg(windows)'.dependencies]
 winreg = "0.55"
 
-[dev-dependencies]
-tauri = { version = "2", features = ["test"] }
-
 [profile.release]
 codegen-units = 1
 lto = true
--- a/src-tauri/src/session_manager.rs
+++ b/src-tauri/src/session_manager.rs
@@ -1193,6 +1193,31 @@ impl SessionManager {
                 ))
     }
 
+    /// Select the active interjection target (backend, app session id, turn id,
+    /// optional ACP client) from a live session, validating that a streaming
+    /// turn is in progress.
+    ///
+    /// Pure (no `AppHandle`) so the rejection path is unit-testable on all
+    /// platforms, including Windows where `tauri::test::mock_app()` crashes the
+    /// test binary with `STATUS_ENTRYPOINT_NOT_FOUND` (tauri #14580 / #13419).
+    fn pick_interjection_target(
+        s: &LiveSession,
+    ) -> Result<(String, String, String, Option<Arc<AcpClient>>), String> {
+        if !(s.prompt_in_flight || s.fsm.state() == SessionState::Streaming) {
+            return Err("interjection requires a streaming turn".into());
+        }
+        let turn_id = s
+            .active_turn_id
+            .clone()
+            .ok_or("interjection requires an active turn")?;
+        Ok((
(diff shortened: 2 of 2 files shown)


shiaho777:
Closing — the mock_app removal did not fix the crash (verified via CI: same `STATUS_ENTRYPOINT_NOT_FOUND` after removal). The real root cause is a test-binary DLL dependency introduced in the mirror-host window. Will reopen with the correct fix + CI diagnostics.

PR #170, opened 2026-07-28
chore(ci): diagnose Windows test-binary DLL load failure

shiaho777:
Diagnostic-only PR (do not merge). Adds a temporary Windows CI step that runs `dumpbin /dependents` on the test binary and attempts a direct invocation, to identify the exact missing DLL behind the `STATUS_ENTRYPOINT_NOT_FOUND (0xc0000139)` crash that has broken every main push since 2026-07-26.\n\nWill be closed once the root cause is identified; the real fix will come in a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,3 +63,26 @@ jobs:
       - name: cargo test
         working-directory: src-tauri
         run: cargo test
+      # ── DIAGNOSTIC (temporary): find the missing DLL behind the Windows
+      # STATUS_ENTRYPOINT_NOT_FOUND (0xc0000139) test-binary crash. Remove
+      # once the root cause is identified and fixed. ──────────────────────
+      - name: Diagnose Windows test-binary DLL deps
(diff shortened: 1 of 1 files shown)


shiaho777:
Moving diagnosis to the fork's own CI to avoid consuming upstream Actions. Will open a clean fix PR once the root cause is identified.

RongleCat:
Thanks for the deep dive — the diagnostic trail (PATH-shadowed API-set DLLs from Temurin JDK → runner image → rustc 1.97.x test binary) is very useful.

Per your note, diagnosis is moving to the fork’s own CI, and this branch still ends with **Rust Windows red** (`STATUS_ENTRYPOINT_NOT_FOUND` / `0xc0000139`) even after the PATH prune + 1.96.0 pin attempt. Closing this diagnostic PR so we don’t land partial CI changes on `main`.

Please open a **clean fix PR** once the root cause is confirmed (happy to review quickly — unblocking Windows CI is high priority). Feel free to link back here.

PR #178, opened 2026-07-28
feat(context): surface agent-reported usage on context chip

sonnemusk:
## Summary
- Host parses ACP usage / token payloads (`usage`, nested `usage`, snake/camel fields)
- Emits `session://usage` for focused and background sessions
- Context chip shows known **input / output / total** when the agent reports them
- Keeps honest `~` estimates when no usage signal exists (never invents zeros)

## Milestone
- M4 / CTX

## Test plan
- [x] `vitest` `src/lib/contextUsage.test.ts` (18)
- [x] `cargo test --lib usage_parse_tests` (3)
- [ ] Manual: after a turn that reports usage, chip loses `~` and menu shows From agent rows
- [ ] Manual: without usage signal, chip still shows `~` estimate + role breakdown

## Out of scope
- Billing currency / cost estimates
- Fake capacity remaining without model context window size


RongleCat:
Integrated on main via local merge (CHANGELOG + multi-PR batch). Thanks!

PR #203, opened 2026-07-29
feat(sidebar): multi-select archive sessions

sonnemusk:
## Summary
- Sidebar **Select** mode: checkbox rows, row click toggles selection (does not open chat)
- Toolbar **Archive N** / **Cancel** with in-app confirm (`setAppDialog`); bulk `sessionSetArchived` via `Promise.allSettled`
- i18n en / zh / zh-TW; light CSS; pure `toggleIdInSet` helper + tests
- Single-session pin/archive/rename menus unchanged; projects not archived

## Test plan
- [ ] With several chats, click **Select** next to Projects
- [ ] Toggle rows via checkbox / row click; **Archive N** confirms and archives; toast shows count
- [ ] Esc / Cancel exits select mode; selection clears after archive
- [ ] Single-session archive, pin, rename, context menu still work outside select mode
- [ ] `pnpm typecheck` and `pnpm test src/lib/sessionSelect.test.ts src/i18n/messages.test.ts`

PR #204, opened 2026-07-29
feat(composer): draft character and word count

sonnemusk:
## Summary

- Show a muted **character + word count** next to the composer when the draft is non-empty
- Pure `draftStats` helpers + unit tests (Unicode-safe char count; whitespace word split)
- Optional Settings → General → Composer toggle (localStorage; **on by default**)
- i18n (en / zh / zh-TW) and settings catalog search entry
- CHANGELOG Unreleased note

## Test plan

- [ ] Type in the composer → muted `N words · M chars` appears; clear draft → counter hides
- [ ] Settings → General → Composer → turn off “Draft character count” → counter gone; turn on → returns
- [ ] Preference survives reload / same-tab settings change via event
- [ ] `pnpm test -- src/lib/draftStats.test.ts src/lib/settingsCatalog.test.ts src/i18n/messages.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ See `docs/llm-wiki/release.md`.
 ### Added
 
 - Desktop notification when the agent asks a question (`session://ask_user`), gated by the existing “Notify on permission” setting
+<<<<<<< ours
 - **Shortcuts catalog**: list **Find in conversation** (⌘/Ctrl+F) and **voice dictation** (Ctrl+Space; not Cmd)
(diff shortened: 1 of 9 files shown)


RongleCat (approved):
Approved after safe re-apply onto main (PR tip had accidental deletions of prior features).

RongleCat:
Integrated onto `main` at `2c6b391` (force batch).

**Note:** The PR tip rebased onto main but rewrote `App.tsx` / settings in a way that dropped sibling features from #197–#203. Maintainer re-applied only the intended additive changes (draft stats / relative timestamps) so those land without regressions.

Thanks @sonnemusk!

PR #205, opened 2026-07-29
feat(appearance): relative message timestamps option

sonnemusk:
## What
Message timestamps can be **absolute** (weekday + clock, existing `formatMessageTime`) or **relative** (“2 minutes ago”, `formatRelativeTime`). Preference is localStorage-only under Appearance → Interface, next to the show-timestamps toggle. Relative labels re-render about every 60 seconds.

## Why
Absolute weekday+time is dense for recent turns; relative times make recency easier to scan without leaving the chat.

## Notes
- Pref: `grok.messageTimeFormat` (`absolute` | `relative`, default absolute)
- Settings search catalog entry `appearance.messageTimeFormat`
- i18n en / zh / zh-TW for format labels and tip
- CHANGELOG Unreleased entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ See `docs/llm-wiki/release.md`.
 ### Added
 
 - Desktop notification when the agent asks a question (`session://ask_user`), gated by the existing “Notify on permission” setting
+<<<<<<< ours
 - **Shortcuts catalog**: list **Find in conversation** (⌘/Ctrl+F) and **voice dictation** (Ctrl+Space; not Cmd)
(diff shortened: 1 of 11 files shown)


RongleCat (approved):
Approved after safe re-apply onto main (PR tip had accidental deletions of prior features).

RongleCat:
Integrated onto `main` at `2c6b391` (force batch).

**Note:** The PR tip rebased onto main but rewrote `App.tsx` / settings in a way that dropped sibling features from #197–#203. Maintainer re-applied only the intended additive changes (draft stats / relative timestamps) so those land without regressions.

Thanks @sonnemusk!

PR #208, opened 2026-07-29
feat(appearance): relative message timestamps option

sonnemusk:
## What
Message timestamps can be **absolute** (weekday + clock, existing `formatMessageTime`) or **relative** (“2 minutes ago”, `formatRelativeTime`). Preference is localStorage-only under Appearance → Interface, next to the show-timestamps toggle. Relative labels re-render about every 60 seconds.

## Why
Absolute weekday+time is dense for recent turns; relative times make recency easier to scan without leaving the chat.

## Notes
- Pref: `grok.messageTimeFormat` (`absolute` | `relative`, default absolute)
- Settings search catalog entry `appearance.messageTimeFormat`
- i18n en / zh / zh-TW for format labels and tip
- CHANGELOG Unreleased entry
- Rebased cleanly on latest `main` (no conflict markers)


sonnemusk:
Already on main via integrate commit 2c6b391 (same as #205). Closing duplicate.

PR #229, opened 2026-07-30
feat(composer): edit items in send queue

sonnemusk:
## What

Adds an **Edit** control on each composer send-queue row so you can change the text of a follow-up before it auto-sends. Save/Cancel go through an in-app GlassModal (textarea), not `window.prompt`. Empty text is blocked unless the item still has attachments.

## Why

The queue already supported guide/remove, but not edit — same gap vs Grok Build TUI `[edit]`. Users often tweak a queued line after steering or while a turn is still running.

## Test plan

- [ ] Queue a follow-up while a turn is streaming; confirm **Edit** opens a dialog with the current text
- [ ] Change text → Save; strip preview updates and item stays in order
- [ ] Save empty text with no attachments → toast error, dialog stays open
- [ ] Save empty text when item has attachments only → allowed
- [ ] Cancel / overlay close → no change; auto-flush can resume after dialog closes
- [ ] `pnpm typecheck` and `pnpm test -- src/lib/sendQueue` pass


RongleCat:
### Integrated on main

PR #230, opened 2026-07-30
feat(context): richer compact dialog with current usage

sonnemusk:
## Summary
- Compact confirm dialog now shows **current context usage** (known / estimated / unknown) from `resolveContextUsageDisplay`
- Short explanation of what compact does + that the optional note is kept for the agent
- Optional **keep-note chips** (decisions / errors / files / TODOs); empty note still allowed
- **Disable Compact** while session is streaming or awaiting permission
- Store UI-side `tokensBefore` at submit; if the agent omits it on the compact event, fill it so the banner can still show `{before} → {after}`
- Unified slash `/compact` path to the same dialog (no `window.prompt` / second API)
- i18n (en / zh / zh-TW) + unit tests for pure helpers + CHANGELOG

## Test plan
- [ ] Open Compact from context chip while idle — dialog shows current usage, explanation, note field + chips
- [ ] Submit with empty note → agent gets `/compact`
- [ ] Submit with chip or custom note → `/compact {note}`
- [ ] While session is streaming, Compact button disabled + busy hint
- [ ] After successful compact, banner shows token range when after is known (even if agent omitted before)
- [ ] Slash command `/compact` opens the same richer dialog
- [ ] `pnpm typecheck` + `pnpm test -- src/lib/contextUsage.test.ts src/i18n/messages.test.ts`


RongleCat:
### Integrated on main

PR #234, opened 2026-07-30
feat(context): system/tools/history breakdown in usage chip

sonnemusk:
## Summary

Fuller **context usage chip** breakdown so the menu can show **System / Tools / History** alongside the existing user / assistant / thought split.

- **Estimation**: classify journal rows into user, assistant, thought, tools (`tool_step` / tool / activity), and system-like; `historyTokens` is the conversation rollup (user+assistant+thought, not double-counted in total).
- **Known agent fields**: when ACP usage payloads include `systemTokens` / `toolsTokens` / `historyTokens`, prefer those numbers **without** `~`.
- **UI**: always show System / Tools / History rows (`—` when unknown / empty estimate; `~` only for estimated non-zero; no tilde for agent-known buckets).
- **i18n**: en / zh / zh-TW keys for the three new labels.
- **Host**: Rust usage parse + `session://usage` emit pass through the optional buckets when present (do not invent zeros).

## Test plan

- [x] `npm run typecheck`
- [x] `npm test -- --run src/lib/contextUsage.test.ts` (24 tests)
- [ ] Manual: open context chip menu on a multi-turn chat with tool activity — expect System/Tools/History + role rows with honest `~`
- [ ] Manual: after agent usage event with structured buckets (if available) — those rows without `~`


RongleCat:
### Integrated on main

PR #240, opened 2026-07-30
feat(trace): keep recent session trace export history

sonnemusk:
## Summary
- After a successful session trace export (`grok trace --local`), record `{sessionId, title?, path, exportedAt}` in a localStorage ring buffer (max ~20, paths only — never file contents).
- Productized **Traces** center: session menu, command palette, and Settings → Runtime → Diagnostics list recent exports with **Reveal in folder** and **Copy path**.
- Pure helper + unit tests (`traceHistory.ts`); i18n en/zh/zh-TW; CHANGELOG.

## Test plan
- [ ] Export a session trace from the chat menu → toast success and entry appears under Traces
- [ ] Open Traces from session menu / command palette / Settings diagnostics
- [ ] Empty state when history is empty
- [ ] Reveal opens Finder/Explorer; Copy path works
- [ ] Re-export past max 20 keeps only the newest 20
- [ ] `pnpm test -- src/lib/traceHistory.test.ts` passes


RongleCat:
### Integrated on main

PR #344, opened 2026-07-31
feat(settings): Ask User Question auto-dismiss timeout

sonnemusk:
## Summary

Adds an **App-enforced** timeout for the `_x.ai/ask_user_question` questionnaire modal, parallel to the existing permission auto-deny timeout.

**Settings → General → Permissions → Ask-question timeout**: Off / 30s / 1m / 2m / 5m (free non-negative seconds also accepted in storage, clamped to 1h).

When enabled, `AskUserModal` shows a countdown and auto-dismisses via the same cancel path as **Dismiss** (`session_resolve_ask_user` → cancelled).

### CLI 0.2.117 alignment

Grok Build CLI **0.2.117** exposes:

```toml
[toolset.ask_user_question]
timeout_enabled = true
timeout_secs = 1800
```

(and env `GROK_ASK_USER_QUESTION_TIMEOUT_ENABLED` / `GROK_ASK_USER_QUESTION_TIMEOUT_SECS` / managed keys `ask_user_question_timeout_*`).

This PR is **App-enforced** (localStorage `grok.askUserTimeoutSec`) — it does **not** rewrite `~/.grok/config.toml` or agent-home. App presets (≤5m) are intentionally shorter UX budgets than the CLI default (~30m). Optional config mirror can land later if product wants dual-write for independent `GROK_HOME`.

## Changes

- `src/lib/askUserTimeout.ts` + pure helper tests
- Settings UI + `settingsCatalog` search entry
- `AskUserModal` countdown + auto-cancel (stable cancel ref so parent re-renders do not reset the timer)
- i18n: en / zh / zh-TW
- CHANGELOG Unreleased note

## Test plan

- [x] `pnpm test -- src/lib/askUserTimeout.test.ts src/lib/settingsCatalog.test.ts src/i18n/messages.test.ts`
- [ ] Settings → Permissions: set Ask-question timeout to 30s; trigger agent questionnaire; confirm countdown + auto-dismiss cancels the gate
- [ ] Off: questionnaire stays open until Submit / Dismiss
- [ ] Quick-pick single-select still answers immediately before timeout

PR #352, opened 2026-07-31
feat(session): stream structured JSON validation + export

sonnemusk:
## Summary

Deepens the existing session **JSON Schema** structured-output panel so turns can be validated **while streaming** (and offline-replayed via pure helpers), with copy/export and optional known usage — no second runtime.

- **Pure helpers** (`src/lib/streamJsonPipe.ts`): stream frames, progressive `assessStreamStructured`, validation status timeline (append + replay), export payload, known-usage pick (never invent zeros)
- **UI**: `StructuredJsonPanel` shows progressive status mid-stream (partial keys, validation path chips), pretty JSON + schema check when complete, copy/export, usage line when agent events reported tokens
- **Wire-up**: existing `sessionJsonSchema` + `jsonSchema` validate path only; ConversationThread mounts panel during stream; App passes labels + `contextUsage.knownUsage`
- **i18n** en / zh / zh-TW · **CHANGELOG** · unit tests

## Test plan

- [x] `pnpm exec vitest run src/lib/streamJsonPipe.test.ts src/lib/jsonSchema.test.ts src/i18n/messages.test.ts`
- [ ] Apply a session JSON Schema, send a prompt that streams object JSON — panel shows Partial → Valid (or schema mismatch) without waiting for turn end
- [ ] Copy / Export JSON when complete; export downloads `structured-output.json`
- [ ] With known `session://usage` events, latest assistant structured panel shows usage detail; missing usage stays hidden
- [ ] Invalid finished reply still shows honest “Not valid JSON” (no crash)
- [ ] Without schema active, structured panel does not appear


RongleCat:
Thanks @sonnemusk — integrated on `main` via batch land (#344–#359). Appreciate the CLI 0.2.117+ surface coverage (settings, i18n, tests).

PR #356, opened 2026-07-31
feat(changes): accept / reject / restore file and hunk diffs

sonnemusk:
## Summary

Session Changes already reviews diffs. This PR adds **Accept / Reject / Restore** per file (and per hunk when before/after snapshots exist).

- **Accept** — keep working tree (write after snapshot when needed)
- **Reject** — restore HEAD via path-scoped `git checkout` (or before snapshot / delete untracked with **in-app confirm** — never wipe untracked without confirm)
- **Restore** — re-apply saved after content after a reject
- Soft-fails when git is missing or project is not a repo
- Pure helpers + host commands are project-path scoped only

## Implementation

| Layer | Detail |
|-------|--------|
| Pure | `src/lib/diffAccept.ts` — parse/apply/reverse unified hunks, reject/accept plans |
| Host | `apply_file_patch`, `git_checkout_file`, `delete_project_file` (under project only) |
| UI | ResourceViewer Changes toolbar + list actions; GlassModal reject confirm |
| i18n | en / zh / zh-TW |
| Tests | vitest `diffAccept.test.ts`; cargo path escape + relative resolve |

## Test plan

- [x] `pnpm exec vitest run src/lib/diffAccept.test.ts src/i18n/messages.test.ts`
- [x] `cargo test --lib git_status_parse_tests::resolve_path`
- [x] `pnpm exec tsc --noEmit`
- [ ] Manual: open Changes on a dirty git file → Accept / Reject (confirm) / Restore
- [ ] Manual: untracked file Reject requires wipe confirm and does not delete without it
- [ ] Manual: non-git project soft-fails or uses before snapshot when available


RongleCat:
Thanks @sonnemusk — integrated on `main` via batch land (#344–#359). Appreciate the CLI 0.2.117+ surface coverage (settings, i18n, tests).

PR #391, opened 2026-07-31
feat(export): share-card pro — honest skins, preview, errors

sonnemusk:
## Summary

EXPORT-SHARE-PRO polish for conversation **Export as image** (share-card PNG):

- **Honest preview stamp**: Save/Copy only enable when the rendered blob matches the selected **session + skin + smart/full** options (no stale PNG after theme/mode switch).
- **Meta chips** under the skin picker: mode · skin · structural layout (smart) · PNG size from the stamp (never invented).
- **Classified errors** for empty / rasterize / load / save / clipboard / cancel (silent on cancel) with en/zh/zh-TW copy instead of raw `Error:` dumps.
- Pure `exportSharePro` helpers + unit tests; CSS polish for skin swatches / error / busy preview states.

## Test plan

- [x] `pnpm exec vitest run src/lib/exportSharePro.test.ts src/i18n/messages.test.ts src/lib/sessionExportImage.test.ts src/lib/shareCardSmart.test.ts`
- [x] `pnpm exec tsc -b`
- [ ] Manual: open Export as image → switch skins / smart toggle → preview rebuilds; Save disabled until ready; empty session shows empty copy (not stack dump); cancel native save stays silent

PR #398, opened 2026-07-31
feat(settings): memory browser pro — search, empty states, honesty

sonnemusk:
## Summary

Polish **Settings → Agent → Memory browser** (alongside memory embedding honesty):

- **Kind chips after content search** — `buildMemoryBrowserDisplayRows` re-applies kind filter after host content-search merge (regression fix: chips previously only affected the client name filter / empty-state check, not the displayed list).
- **Contextual empty states** — pure `resolveMemoryBrowserEmptyState` for off / loading / empty catalog / searching / no matches / filtered, with mode-aware hints and **Clear filters**.
- **Honesty** — App search stays **keyword-only** (never invents embeddings); empty-search hints + soft-link to Memory embedding when `embedding.model` is unset; name vs **content** match badges + match summary.
- Pure helpers + Vitest; en / zh / zh-TW; `settingsCatalog`; CHANGELOG.

## Test plan

- [x] `npx vitest run src/lib/memoryBrowserSearch.test.ts src/lib/memoryBrowserFilter.test.ts src/lib/memoryEmbedConfig.test.ts src/i18n/messages.test.ts`
- [x] `npx tsc --noEmit`
- [ ] Settings → Agent → Memory browser: kind chip filters the list with and without a search query
- [ ] Empty catalog / filter empty / no search matches show distinct hints; Clear filters works
- [ ] With embedding model unset, no-match empty offers link to Memory embedding settings
- [ ] Content hits show “content” badge; name-only matches show “name”; match summary updates


RongleCat:
Thanks @sonnemusk — landed on `main` via batch integrate #391–#399.

PR #399, opened 2026-07-31
feat(remote-im): add native Telegram controls and account switching

KAMIENDER:
## Summary

- register Telegram native slash commands for the Remote IM bot
- render project, session, and account results as Telegram inline keyboards
- support same-message pagination with previous/next controls and text-number fallback
- show account quota details and validate saved account IDs before switching
- preserve callback ACL checks, acknowledge callbacks promptly, and document the Telegram flow

## Why

Telegram users currently have to read numbered text results and reply manually. Native menus and inline selection make project switching, session recovery, and account switching easier to discover and operate, especially when result sets are large.

## Impact

- Telegram Remote IM gains native command discovery and interactive selection
- project/session/account lists are paginated at 20 entries per page
- existing text commands and numbered selection remain available as a fallback
- account switching reconnects the desktop ACP session through the existing account-change flow

## Validation

- `pnpm build:mac-arm`
- installed the generated ARM64 app bundle locally
- confirmed the relaunched app established the Telegram long-poll connection

No separate unit test suite was run, per the repository-local instruction.


RongleCat:
Thanks @KAMIENDER — landed on `main` via batch integrate #391–#399.

PR #493, opened 2026-08-01
feat(settings): CLI sessions search empty honesty

sonnemusk:
## Summary
- Pure `cliSessionsSearchPro` helpers: empty honesty (`loading` · `searching` · `cli_missing` soft-fail · empty catalog · filter/search empty + clear CTA), link filter chips (all · linked · unlinked), ranked free-text hits (title → id → first prompt → cwd), classified search/list errors, `planImportSelection` honesty counts — **never invents CLI sessions**.
- Settings → CLI sessions UI: filter chips with counts, error chips (soft vs hard), clear-filters CTA, import/delete buttons use honest counts (skip already linked; delete only on-disk unlinked). Delete stays on GlassModal (no `window.confirm`).
- i18n en/zh/zh-TW, `settingsCatalog`, CHANGELOG.

## Verify
```
pnpm test -- src/lib/cliSessionsSearchPro.test.ts src/lib/cliSessionsSearch.test.ts src/i18n/messages.test.ts
pnpm typecheck
```

## Test plan
- [ ] Open Settings → Agent / CLI sessions with CLI present: list loads; link chips filter; search ranks title hits first
- [ ] Clear filters CTA when filter/search empty
- [ ] Soft-fail empty when CLI missing and list empty (no invented rows)
- [ ] Search host failure shows soft error chip + local filter fallback
- [ ] Import all / delete unlinked counts match honesty plan; delete still uses GlassModal

PR #454, opened 2026-08-01
feat(changes): comment on diff hunk into composer

sonnemusk:
## Summary
- Adds per-hunk **Comment** on the Changes panel: GlassModal for a free-form review note, then inserts a structured prompt (file path, hunk header/snippet, note) into the chat **composer** (draft insert, not auto-send).
- Pure helpers in `src/lib/diffComment.ts` (`formatHunkSnippet`, `validateDiffCommentNote`, `buildDiffCommentPrompt`, `planDiffCommentToChat`) with NUL strip, secret redact, and ~4k caps.
- Parent `App` sets draft + focuses composer + soft success toast; button hidden when callback is omitted.
- i18n en / zh / zh-TW; CHANGELOG under Composer & chat; no `window.confirm`.

## Test plan
- [x] `pnpm test -- src/lib/diffComment.test.ts src/i18n/messages.test.ts`
- [x] `pnpm typecheck`
- [ ] Manual: open Changes with a multi-hunk file → **Comment** on a hunk → enter note → **Insert into chat** → composer shows structured prompt and is focused; toast appears; Send not auto-fired.
- [ ] Manual: empty note shows validation error; cancel closes modal without changing draft.
- [ ] Manual: accept/reject hunk buttons still work; comment button absent if `onDiffCommentToChat` not wired (not applicable in App).

PR #494, opened 2026-08-01
fix(ci): sync pnpm-lock for @vitejs/plugin-react-swc

sonnemusk:
## Summary

- Sync `pnpm-lock.yaml` with `package.json` after `@vitejs/plugin-react-swc` was added
- Restores `pnpm install --frozen-lockfile` on main (and PR merge CI)

## Test plan

- [x] `pnpm install --frozen-lockfile` succeeds locally
- [ ] CI frontend job green

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -126,6 +126,9 @@ importers:
       '@vitejs/plugin-react':
         specifier: ^4.5.2
         version: 4.7.0(vite@6.4.3(@types/node@22.20.1)(jiti@2.7.0)(lightningcss@1.32.0))
+      '@vitejs/plugin-react-swc':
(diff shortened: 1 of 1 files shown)


sonnemusk:
## Status (rebased after #503)

- Rebased onto current `main` (`145f7535` — includes integrate #454–#493).
- Root cause of frontend CI red: `package.json` has `@vitejs/plugin-react-swc` but `pnpm-lock.yaml` does not → `pnpm install --frozen-lockfile` fails.
- Local verify: `pnpm install --frozen-lockfile` succeeds on this tip (`740efca8`).
- **Please squash-merge this ASAP** so main frontend CI is green again after #503.

This is the only remaining tail of the 8/1 startup optim (SWC switch).

PR #505, opened 2026-08-01
perf(chat): stream isolation for multi-turn jank (Intel Retina)

sonnemusk:
## Summary

Ports and lands a multi-turn chat performance suite on current `main` (post AppWorkbench split / v0.2.4), aimed at freezes after a few agent turns on dual-GPU Intel Retina laptops (e.g. 2019 16″ MBP).

### Hot path
- **External stores**: transcript (`sessionTranscriptStore`), multi-session live map (`sessionLiveMapStore`), focused session + liveHost (`sessionShellStore`)
- **`useSessionRuntime`** wired through those stores; shell keeps a **structural** messages mirror so token growth does not re-render the whole workbench
- **`ConversationThreadLive`** subscribes to full transcript content; workbench no longer passes `messages={…}` into the thread
- Stream coalesce flush is **hardware-adaptive**; long streaming bodies use plain pre-wrap then one markdown pass on settle
- Virtualize earlier; shrink overscan under `data-stream-perf`; throttle stick-to-bottom ResizeObserver while streaming

### UI / GPU
- Stream-perf mode dims wallpaper, hides wallpaper video, drops glass `backdrop-filter` during active turns
- Memoized transcript rows + stable path-map identity (siblings skip re-render when only the tail grows)
- Relative timestamps still refresh via `timeTick`

### Bundle
- Lazy `CodePreview` / Office / Tiptap / `FileMediaPlayer` in `ResourcePreviewBody`
- Default transcript filter conversation-only (fewer tool DOM rows)

### Workbench
- Live Cmd/Ctrl+F re-reads the transcript store while find is open (structural shell mirror stays cheap)

## Test plan

- [x] `tsc -b`
- [x] Vitest: stream / store / virtual list / transcript filter suites (75 tests)
- [ ] Manual: multi-turn tool-heavy chat on Intel Retina — stream should not freeze the shell
- [ ] Manual: wallpaper on → stream — wallpaper/video should drop load mid-turn
- [ ] Manual: Cmd/Ctrl+F during stream finds current content
- [ ] Manual: open code/office preview in resource pane (lazy load)

### DIFF
--- a/src/app/AppWorkbench.tsx
+++ b/src/app/AppWorkbench.tsx
@@ -918,6 +918,7 @@ import {
 import type { ContextMenuState } from "@/lib/app/appDialogTypes";
 import { useSessionRuntime } from "@/hooks/useSessionRuntime";
 import { sessionTranscriptStore } from "@/lib/sessionTranscriptStore";
+import { useLiveMapWhen } from "@/hooks/useSessionLiveMap";
 import { useComposerController } from "@/hooks/useComposerController";
 import { useAppDialogs } from "@/hooks/useAppDialogs";
 import { useSessionHostEvents } from "@/hooks/useSessionHostEvents";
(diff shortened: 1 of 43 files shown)


RongleCat (approved):
Integrated onto main after resolving conflicts with UserAttachments / ChatCut handoff / stick-to-bottom. CI was green; local typecheck + store/virtualizer tests pass.

PR #508, opened 2026-08-02
perf(markdown): lazy highlight.js + plain stream body

sonnemusk:
## Summary
Secondary code-splitting for the markdown/code path so chat cold start and long streaming turns do not pay full highlight.js / remark cost up front.

- **CodePreview**: dynamic `import()` of highlight.js core + languages (cached); plain escaped text until ready
- **ResourceViewer**: `React.lazy` CodePreview with plain-code Suspense fallback
- **MarkdownChat**: plain pre-wrap body while streaming past 2000 chars; throttle ReactMarkdown re-parse (~160ms); lazy `CodeBlock`

## Test plan
- [ ] Open a code file in the resource pane — first open may load hljs chunk; content still readable immediately
- [ ] Stream a long assistant reply with code fences — UI stays responsive; settles to full markdown when done
- [ ] `pnpm exec tsc --noEmit`
- [ ] `pnpm exec vitest run src/lib/streamRenderPolicy.test.ts`

Independent of #505 stream isolation PR.

### DIFF
--- a/src/components/CodePreview.tsx
+++ b/src/components/CodePreview.tsx
@@ -1,40 +1,13 @@
 /**
  * Resource-pane code preview — highlight.js (same stack as Grok Desktop)
  * with light/dark themes bound to `data-theme` on documentElement.
+ *
+ * highlight.js core + languages load on first mount via a cached dynamic
+ * import so ResourceViewer / chat cold path never pay language-pack cost at
+ * module eval. While loading, plain escaped text is shown immediately.
  */
 
 import { useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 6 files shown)


RongleCat:
Closing as **largely landed via #505**.

`streamRenderPolicy` + streaming plain markdown / lazy code path already exist on main after the stream-isolation merge. Remaining hunks no longer apply cleanly (MarkdownChat / ResourcePreviewBody conflict with the #505 ports).

If any highlight.js lazy edge is still missing vs this PR, please open a small follow-up against current main.

Thanks @sonnemusk!

PR #510, opened 2026-08-02
perf(composer): external draft store + memo editor

sonnemusk:
## Summary
Stop every keystroke from re-rendering the App monolith, and let stream-driven App re-renders skip the contenteditable when editor props are stable.

- **composerDraftStore** + hooks — draft lives outside App state; App uses `getDraft()` / `setDraft` without subscribing
- **ComposerDraftEditor / Chrome islands** — only small trees subscribe to draft content/meta
- **memo(ComposerEditor)** — skip re-render when value/disabled/placeholder unchanged during stream

## Test plan
- [ ] Type in composer during a streaming turn — input stays smooth; no focus loss
- [ ] IME composition (CJK) still works
- [ ] Send / clear / prompt history / project draft restore behave as before
- [ ] `pnpm typecheck` / `tsc --noEmit`
- [ ] `pnpm exec vitest run src/lib/composerDraftStore.test.ts`

Independent of #505 stream isolation PR.

### DIFF
--- a/src/app/AppWorkbench.tsx
+++ b/src/app/AppWorkbench.tsx
@@ -5,6 +5,7 @@ import {
   useRef,
   useState,
   type CSSProperties,
+  type KeyboardEvent as ReactKeyboardEvent,
   type MouseEvent as ReactMouseEvent
 } from "react";
 import { useThemeShell } from "@/providers/ThemeProvider";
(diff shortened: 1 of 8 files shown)


RongleCat (approved):
LGTM — external composer draft store stops keystroke re-renders of AppWorkbench; CI green and mergeable on current main.

PR #511, opened 2026-08-02
perf(chat): memo row components and virtualize earlier

sonnemusk:
## Summary
Reduce multi-turn chat DOM/render cost on Intel Retina:

- **Memo** hot lobe-chat rows: ChatItem, Thinking, Timeline*, TurnActivity, EndOfTurn, BackBottom
- **Virtualize earlier**: threshold 48 → 16 rows
- **Stream overscan scale**: ~0.55–0.75 while streaming (fewer mounted siblings)

Uses `chatVirtualPolicy.ts` (not `streamRenderPolicy`) to avoid clashing with #508.

## Test plan
- [ ] Multi-turn agent chat with many tool steps still scrolls / pins correctly
- [ ] Streaming tail updates without re-rendering all history rows (React profiler optional)
- [ ] `pnpm typecheck` / virtual list unit tests

Independent of #505 stream isolation.


RongleCat:
Closing as largely landed via #505. CHAT_VIRTUALIZE_THRESHOLD_PERF=16 and resolveStreamOverscanScale already on main; lobe-chat rows are memoized. Rename to chatVirtualPolicy not worth conflict surface. Thanks @sonnemusk!

PR #513, opened 2026-08-02
perf(ui): stream-perf mode cuts wallpaper and backdrop thrash

sonnemusk:
## Summary
While a turn is streaming / busy, set `data-stream-perf=1` on `<html>`:

- Dim / hide wallpaper video decode
- Disable backdrop-filter on glass / sidebar / floating composer
- Throttle stick-to-bottom ResizeObserver (~100ms)

Targets Intel integrated-GPU Retina composite thrash during multi-turn streams.

## Test plan
- [ ] Start a streaming turn — wallpaper dim, blur off; after turn ends, restore
- [ ] Pin-to-bottom still works while streaming
- [ ] `pnpm typecheck`

Independent of #505.


RongleCat:
Integrated onto main (maintainer cherry-pick / conflict resolve after #505). Thanks @sonnemusk!

PR #509, opened 2026-08-02
perf(sidebar): memo session row + isolate relative time

sonnemusk:
## Summary
Isolate sidebar session rows so App re-renders (stream / chrome) do not rebuild every row tree when props are unchanged.

- **SidebarSessionRow** — memoized row for project + orphan VirtualLists
- **SidebarSessionRelativeTime** — 60s tick via `relativeTimeTickStore` (no App-wide `setState` every minute)
- **CSS** — `content-visibility` / `contain` on `.tree-l3`

## Test plan
- [ ] Sidebar sessions list: open, pin, archive, context menu still work
- [ ] Busy spinner still shows on working sessions
- [ ] Relative time labels refresh without janking the whole app
- [ ] `pnpm exec tsc --noEmit`
- [ ] Vitest: `relativeTimeTickStore` + `SidebarSessionRow` tests

Independent of #505 stream isolation PR.

PR #514, opened 2026-08-02
perf(chat): island ConversationThread rows and relative time

sonnemusk:
## Summary
Split ConversationThread so stream / time ticks don't rebuild the whole transcript tree:

- **TranscriptMessageRow** — memoized per-row island with custom equality
- **ChatMessageTime** — relative labels subscribe to a shared 60s tick store (no parent `setState`)
- **ConversationStickyTail** — live tool + quiet thinking under the list
- Thread file shrinks ~1940 → ~988 lines of orchestration

## Test plan
- [ ] Multi-turn chat: streaming updates, find, regenerate, edit last user still work
- [ ] Relative timestamps refresh without janking scroll
- [ ] `pnpm typecheck` / new unit tests

Independent of #505; shares `relativeTimeTickStore` idea with #509 (comment-only diff if both land).

### DIFF
--- a/src/components/lobe-chat/ChatMessageTime.test.tsx
+++ b/src/components/lobe-chat/ChatMessageTime.test.tsx
@@ -1,0 +1,53 @@
+import { describe, expect, it } from "vitest";
+import React from "react";
+import { renderToString } from "react-dom/server";
+import { ChatMessageTime } from "./ChatMessageTime";
+
+describe("ChatMessageTime", () => {
+  const createdAt = "2026-01-15T12:00:00.000Z";
+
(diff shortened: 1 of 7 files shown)


RongleCat:
Closing as **largely landed** via #505 (inline `TranscriptMessageRow` islands in ConversationThread) + #509 (`relativeTimeTickStore`).

Extracting the row into a separate file / ConversationStickyTail would still be a nice structural follow-up against current main, but replaying this PR on post-#505 ConversationThread is high-conflict for limited net gain.

Thanks @sonnemusk — open a slim extract PR if you want the file split without re-landing isolation.

PR #532, opened 2026-08-07
fix/feat: side pane close, titlebar maximize, mark as unread

enderzcx:
## Summary

Three desktop UX items from feedback:

### 1. Right side pane opens but cannot close (non-maximized)
- Sync-clamp aside width before first paint; keep main top toggle while open
- Re-clamp window position after grow; protect side chrome under narrow width

### 2. Double-click top chrome does not fill work area (mac)
- Enable titlebar maximize on all desktop hosts (was Win-only)
- `mousedown(detail=2)` fallback when drag regions swallow `dblclick`
- Track `is-maximized` on mac; wire sidebar / main / side tab / settings / setup

### 3. No **Mark as unread** on session menu
- Toggle: already unread → **Mark as read**; else → **Mark as unread**
- Uses existing `markUnread()` storage; sidebar dots + dock badge update immediately
- Manual mark on the **currently open** chat holds until leave + re-open (so focus/visibility auto-clear does not wipe it instantly)
- **Clear all unread** still available when any unread exists

## Test plan

- [ ] Windowed: open right pane → close via main top + side chrome + shortcut
- [ ] mac: double-click left sidebar top / main top empty → maximize / restore
- [ ] Right-click a **read** session → **标为未读** / Mark as unread → blue unread dot
- [ ] Right-click that session → **标为已读** / Mark as read → dot clears
- [ ] Mark **current** chat unread → switch away → switch back → unread clears on re-open
- [ ] Clear all unread still works
- [ ] `pnpm exec vitest run src/lib/sessionUnread.test.ts src/i18n/messages.test.ts`

PR #533, opened 2026-08-07
fix(composer): persist per-session input drafts across thread switches

enderzcx:
## Summary

- Switching between existing threads no longer drops a half-typed follow-up in the composer.
- Adds per-session draft storage (`localStorage` key `grok.composerSessionDrafts`): text, attachments, and goal mode.
- Save when leaving a real session, restore immediately when opening it, debounced persist while typing, clear on send or explicit clear.
- New-chat continues to use the existing per-project draft buffer; the two paths stay separate.

## Why

`openSession` always cleared the composer for real sessions, and only new-chat pages stashed drafts under the project key. Users typing a follow-up, switching threads, and coming back lost their text.

## Test plan

- [x] Unit tests: `src/lib/composerSessionDraft.test.ts` (9 cases) + existing project draft tests
- [x] Manual: type in thread A → switch to B → back to A → draft restored
- [x] Manual: send from A → switch away and back → composer empty
- [x] Manual: new-chat project buffer still restores independently
- [ ] CI green

## Notes

Local smoke verified on a rebuilt `Grok.app` after a full process restart (replacing the binary while the old process was still running does not pick up the fix until relaunch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Fixed
+- **Per-session composer drafts**: Switching threads no longer drops a half-typed follow-up. Each real session keeps text / attachments / goal mode in `localStorage` (`grok.composerSessionDrafts`); restore on open, debounced persist while typing, clear on send or explicit clear. New-chat still uses the existing per-project buffer.
+
+**中文 · 修复**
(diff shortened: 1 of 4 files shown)


RongleCat (approved):
## Review

**Verdict: Approve + squash-merge**

Real sessions only cleared the composer on open and never stashed follow-ups, so switching threads dropped half-typed prompts. This mirrors the existing per-project new-chat buffer with a per-session `localStorage` map:

- save on leave / debounced while typing
- restore on open (before journal load)
- clear on send / explicit clear
- capped at 120 entries; corrupt storage soft-fails

**Checks**
- [x] Unit tests (`composerSessionDraft` 9 cases) pass
- [x] Rebased onto `main` after #532; CHANGELOG conflict resolved under **Unreleased**
- [x] No secrets / no window.confirm
- [x] Scope matches description

Thanks @enderzcx.

PR #534, opened 2026-08-07
feat(providers): DeepSeek balance probe in settings and sidebar

falser101:
## Summary

- Add Host `providers_balance` for DeepSeek only: `GET https://api.deepseek.com/user/balance` with Bearer API key (amounts kept as strings; `/v1` stripped from stored base).
- Settings → Account → Custom providers: **Check balance** card with availability + total / granted / topped-up lines.
- When DeepSeek is the active custom route: sidebar footer shows `110.00 CNY` style one-liner; UserMenu shows detail + refresh (5-minute session TTL, no invented zeros).
- Extensible adapter shape for future Coding Plan / other provider probes; docs + CHANGELOG + unit tests (Rust + Vitest).

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml deepseek_balance`
- [x] `pnpm exec vitest run src/lib/providerBalanceHonesty.test.ts src/lib/providerBalanceFormat.test.ts`
- [x] `pnpm exec tsc --noEmit`
- [ ] Activate DeepSeek custom provider with a real key → sidebar footer shows balance line
- [ ] Open UserMenu → detail + Refresh updates balance
- [ ] Settings → Check balance shows full `balance_infos` rows
- [ ] Amux / Yun / OpenCode Go / 火山方舟: no balance button / no footer line
- [ ] Bad key: soft-fail auth copy, never shows fake `0.00`


falser101:
图片
图片

PR #535, opened 2026-08-07
fix(media): copy image via native OS clipboard on desktop

enderzcx:
## Summary

- Chat / attachment / lightbox **Copy image** now prefers Host `clipboard_write_image` (arboard) on Tauri, so the image lands on the real OS clipboard.
- WebView `navigator.clipboard.write(ClipboardItem image/png)` often fails silently on macOS WKWebView — paste into Feishu and other apps gets nothing.
- Falls back to the browser clipboard API when not on desktop or if the native path throws.

## Why

Session export already used `api.clipboardWriteImage` for this reason; image cards still went through `copyImageFromSrc` → WebView only.

## Test plan

- [ ] Open a chat with an agent image (e.g. relative `images/*.png` / project preview PNG)
- [ ] Right-click → **Copy image**
- [ ] Paste into Feishu / Preview / Notes — image should appear
- [ ] Attachment card copy image still works
- [ ] Lightbox right-click copy still works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,12 +18,14 @@ See `docs/llm-wiki/release.md`.
 - **DeepSeek 余额查询**：Host `providers_balance` 请求官方 `/user/balance`；设置页可查完整明细，激活 DeepSeek 时侧栏左下角与个人菜单显示 `110.00 CNY` 一行（会话内 5 分钟缓存，失败不编造 0）。
 
 ### Fixed
+- **Copy image to OS clipboard (Tauri)**: Chat / attachment / lightbox “Copy image” now uses Host `clipboard_write_image` (arboard) instead of WebView `ClipboardItem`, so paste works in Feishu and other apps. Falls back to the browser clipboard API when not on desktop.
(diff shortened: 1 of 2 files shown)


RongleCat (approved):
## Review

**Verdict: Approve + squash-merge**

WebView `ClipboardItem` image write is unreliable on macOS WKWebView (silent fail → Feishu/other apps get nothing). Prefer existing Host `clipboard_write_image` (arboard), same path session export already uses, with browser fallback.

**Checks**
- [x] Reuses existing `clipboardWriteImage` / `isTauri` from `@/lib/api` (`export *` extensions)
- [x] Minimal scope (`copyImage.ts` + CHANGELOG only)
- [x] MERGEABLE vs main
- [x] No secrets / no UI chrome issues

Thanks @enderzcx.

enderzcx:
## CUA verification (local desktop)

Reproduced the bug with cua-driver on the installed app: right-click **复制图片** left the clipboard as plain text only (Feishu paste empty).

Follow-up commit strengthens the path:

1. Host `clipboard_write_image_path` (disk → arboard), re-grants path on user-initiated copy
2. Extract abs path from loopback media URLs (`/v1/media?p=…`)
3. Fallback: paint the loaded `` to PNG then Host write
4. Last resort: base64 / browser ClipboardItem

**Retest:** after rebuild/relaunch, same CUA flow → clipboard reports `PNGf` / TIFF and `pngpaste`-style save wrote a 62KB PNG. Ready for human check in Feishu.

PR #547, opened 2026-08-11
fix(tool-timeline): capture ACP output, fix tool identity, free plan mode

shiaho777:
## What

Three coupled defects, verified against real session journals before fixing.

### 1. Tool output was never captured → tools couldn’t expand
`extract_tool_ui_fields` only read `rawInput` (command/query/description), so `detail` was the **call argument echoed back** — never the result. `read_file` / `grep` / `list_dir` produced no expandable body (`hasBody=false` → row rendered as a `<span>`, not a clickable `<button>`).

- New `extract_tool_output` reads ACP `content[]` (text items + diff headers).
- Emit on `session://tool`; persist behind a sentinel so the legacy `tool_step|` body parses byte-identically (old journals unaffected).
- New shared `ToolExpandBody` renders a terminal-transcript block (`$ cmd` above a boxed, internally-scrolling output), used by both `TimelineToolRow` and `GrokActivityStepRow`.

### 2. Machine tool name was unreachable → bare `tool` label
The `_meta` key is literally `"x.ai/tool"`; JSON Pointer requires the inner slash escaped as `~1` (RFC 6901). `/_meta/x.ai/tool/name` never resolved, so on sparse terminal updates the journal collapsed to `tool_step|failed|tool|tool`.

- New `tool_meta()` / `tool_meta_str()` helpers use the escaped pointer.
- Fixed both call sites (`session_manager/types` + `acp_client` permission).

### 3. Plan mode could not be cancelled
`mode: "plan"` was persisted into global settings **and** project defaults, so every new session inherited it; `PlanStatusBar` rendered no exit control for `kind==="plan_mode"`.

- Plan mode is transient/per-session: `sticky_mode` drops it at every global/project persistence site; `non_plan_mode` heals the value already on disk.
- The idle plan chip now has a close button (`exitPlanMode`) that flips back to agent.

## Verification
- `updates.jsonl` (146 real tool events): `toolName` appears 0 times, `kind` only 44 — the broken pointer was the sole identity source on sparse payloads.
- Reproduced the bare-`tool` rows in `sessions/96521dc7…/messages.json` (`tool_step|failed|tool|tool`).

## Tests
+5 Rust (pointer escape, enrich recovery, output extraction, plan-mode healing/stickiness), +5 TS (output expand body, read-file expandability, long-output elision, sentinel parser, output threading + survives sparse tick).

Base of a stack of 3. Merge order: **#547 → #548 → #549**.

### DIFF
--- a/src-tauri/src/acp_client.rs
+++ b/src-tauri/src/acp_client.rs
@@ -3440,7 +3440,8 @@ pub fn decode_permission_request(rpc_id: u64, params: &Value) -> AcpEvent {
     // Kind alone breaks tool-scoped session fallbacks (#542 shell →
     // allow-always-command).
     let tool_name = tool_call
-        .pointer("/_meta/x.ai/tool/name")
+        // `_meta` key is literally "x.ai/tool" — JSON Pointer needs `~1` (RFC 6901).
+        .pointer("/_meta/x.ai~1tool/name")
         .or_else(|| tool_call.pointer("/_meta/tool/name"))
         .or_else(|| tool_call.get("name"))
         .and_then(|v| v.as_str())
--- a/src-tauri/src/cli_sessions.rs
+++ b/src-tauri/src/cli_sessions.rs
@@ -16,7 +16,9 @@ use serde_json::Value;
 use uuid::Uuid;
 
 use crate::paths::resolve_agent_grok_home;
-use crate::session_manager::{extract_tool_input, tool_journal_richer};
+use crate::session_manager::{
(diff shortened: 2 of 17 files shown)


shiaho777:
image

shiaho777:
image

shiaho777:
0.2.14使用计划模式后新建会话会持久存留
image
修复该问题

RongleCat:
Merged to `main` (`02411e17`). Thanks @shiaho777 — solid root-cause work on tool output capture, `x.ai/tool` pointer escape, and plan-mode stickiness.

PR #548, opened 2026-08-11
feat(activity): group mixed read+search bursts into one expandable row

shiaho777:
## What

The phase rail only collapsed **pure-search** runs into “Ran N searches”. A mixed read+search burst rendered as N flat rows. This folds them into one expandable **“Explored · N searches, M files”** header.

- New `GrokActivityStep` variant `explore-group` with `searches`/`reads` counts and child steps.
- Detection runs before the browse/search leaf handlers: a maximal run of read/search context tools (browse excluded — it keeps its “Browsed …” row) of length ≥ 2, **unless** the run is pure-search (existing `search-group` path wins).
- Children built via an internal `grouping=false` recursion so leaves never re-trigger grouping (no infinite loop).
- `TimelinePhaseBlock` renders the group expandable; expanded body nests a `GrokActivitySteps` list — each child keeps its own output expand from #547.
- i18n: `chat.explored` / `exploreSearches` / `exploreFiles` (+singular) for en / zh / zh-TW.

## Tests
+4 explore-group (mixed fold, pure-search excluded, edit breaks run, browse breaks run); updated 1 pre-existing test whose flat 4-row assertion became an explore-group + 2 leaves.

## Stack
Depends on #547 (imports `ToolExpandBody` and the new `toolExpandBody` shape). Contains #547’s commit until it merges. Merge order: **#547 → this → #549**.


shiaho777:
image

RongleCat:
Merged to `main` with the stack (`02411e17`). Thanks @shiaho777 — explore-group for mixed read+search bursts looks good; i18n en/zh/zh-TW covered.

PR #549, opened 2026-08-11
fix(tool-timeline): recover kind on replay + humanize fallback label

shiaho777:
## What

Closes the residual case where internal CLI tools (`enter_plan_mode` / `exit_plan_mode`) and any tool without an `_meta` block still showed the bare generic tool label after the pointer fix in #547.

### 1. Replay lost `toolKind`
`toolSegmentFromMessageRow` read `row.toolKind` (always empty for history-loaded rows) + `inferKindFromToolCallId`. It never fell back to the **journal body’s parsed `kind`**, so replayed tools lost their typed identity and collapsed to fallback. Now parses the body first and uses `parsed.kind` as the fallback.

### 2. Bare generic label for unknown tools
New `humanizeToolKind()`: the fallback bucket shows a readable machine-name label (“Enter Plan Mode”) instead of the bare generic label. Acronyms (`api`/`url`/`mcp`/`x`…) keep casing; the bare generic label is the last resort, only when nothing is known about the tool.

## Verification
Confirmed against the real plan-test session journal (`sessions/96521dc7…`), which has `enter_plan_mode` / `exit_plan_mode` rows that previously rendered as the bare generic label.

## Tests
+4 (`humanizeToolKind` cases, empty/`tool` → undefined, fallback uses humanized name, generic-only-when-unknown) + 1 replay kind-recovery via `weaveToolsIntoAssistantSegments`.

## Stack
Stacked on #547 + #548. Merge order: **#547 → #548 → this**.


RongleCat:
Merged to `main` with the stack (`02411e17`). Thanks @shiaho777 — kind recovery on replay + humanize fallback closes the residual bare-tool label case.

PR #550, opened 2026-08-11
fix(thinking): collapse on episode end like the work block

shiaho777:
## What

The standalone thinking block stayed expanded after the answer began. Root cause was a `keep-open` preference that re-opened finished blocks on every reload/remount, plus a fragile live→done transition that relied on a previous-value ref (lost when the host message id stabilised mid-stream and the component remounted).

Match the work/phase block model: open while the episode is live, fold the instant it ends, independent of any preference.

- `ConversationThread`: the thinking prop is now `m.streaming && !messageHasContent` — true only while the message streams AND no body segment has arrived, so it flips to false the moment the answer starts (not only at turn end).
- `Thinking`: drop the pref-driven open/collapse machinery (pref state, pref listener, `expandPref` prop). Initial open tracks `thinking`; the live-tick effect collapses whenever the episode is not live and the user has not manually expanded. A finished block can still be clicked open.

The keep-open setting in Settings is unchanged, but finished thinking blocks no longer auto-open from it.

## Verification
A temporary `DBG th=… open=… pref=…` chip on the block header confirmed the prop went false (`th=0`) while `open` stayed `1` under `pref=keep-open` — i.e. the pref, not the signal, was holding it open. After the change the block collapses on episode end regardless of pref.

## Stack
Stacked on #547 → #548 → #549. Merge order: …#549 → **this** → #552.


RongleCat:
Merged to `main` with the stack (`02411e17`). Thanks @shiaho777 — thinking collapse parity with the work block is the right model.

PR #551, opened 2026-08-11
feat(activity): make a phase’s thought step expandable to full text

shiaho777:
## What

Inside a “worked for …” phase, the 💡 thought step only rendered its one-line summary with no way to open it — so on reload the reasoning was effectively unreadable while the tool rows alongside it expanded fine.

Make the thought step expandable like the others: when it has body text, the row becomes a button whose expanded body renders the full thought via `MarkdownChat` (muted), capped at 240px with internal scroll so a long reasoning trace does not bury the surrounding tool rows. Thread `locale` through `GrokActivitySteps` / `GrokActivityStepRow` so the markdown renderer can localise.

## Stack
Stacked on #547 → #548 → #549 → #550. Merge order: …#550 → **this**.


shiaho777:
修复的问题是在版本中工作块的思考块及正文会丢失，只显示工具调用过程

RongleCat:
Merged to `main` as stack tip (`02411e17`). Thanks @shiaho777 — expandable phase thought steps fill the last gap next to tool rows.

PR #552, opened 2026-08-11
fix(ask-user): show the questionnaire when rpcId is 0

shiaho777:
## What

The `session://ask_user` listener validated the payload with `!p?.rpcId`, which treated a valid JSON-RPC id of 0 as missing. The agent’s first `_x.ai/ask_user_question` carries rpc id 0, so the modal was silently dropped and the turn hung with the stop button showing “working” until the user cancelled.

Replace the truthy guard with an explicit null check (`p.rpcId != null`). Extracted into `isValidAskUserPayload` so the rpcId=0 regression is unit-tested.

## Reproduced from a real session log
```
acp ask_user_question id=0 questions=1 tool_call=call_00_…
<2 min hang — modal never shown>
acp → ask_user_question reply id=0 outcome=cancelled
```

## Tests
+5 `isValidAskUserPayload` (accepts id 0, rejects missing rpcId / empty questions / null payload).

## Scope
Independent of the tool-timeline / thinking stack (#547–#551); branches off `main`.

### DIFF
--- a/src/hooks/useSessionHostEvents.ts
+++ b/src/hooks/useSessionHostEvents.ts
@@ -6,6 +6,7 @@
 import { useEffect, useRef } from "react";
 import * as api from "@/lib/api";
 import { isMirrorClient } from "@/lib/mirrorTransport";
+import { isValidAskUserPayload } from "@/lib/askUserPayload";
 import {
   applyContextCompact,
   applyGeneratedImage,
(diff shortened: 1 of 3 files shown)


shiaho777:
image

shiaho777:
image
在之前的版本中调用tool_step|completed|ask_user_question|Ask User工具时不显示问询框，导致任务卡住

RongleCat:
Merged to `main` (`be7706c5`). Thanks @shiaho777 — classic rpcId=0 truthy-guard bug; clear repro and unit coverage.

PR #594, opened 2026-08-12
feat(providers): per-model connection test

shiaho777:
## Summary
Add a per-model **Test connection** button in the custom-provider form (Settings → Account → Custom providers). Each model row gets a button that sends one tiny non-streaming inference request (`max_tokens: 1`, prompt `"hi"`) and reports **✓ on HTTP 2xx** or **✗ with a reason** on failure.

- New Tauri command `providers_test_model`.
- Honors the channel's `api_backend` (`chat_completions` → `/chat/completions`, `responses` → `/responses`, `messages` → `/v1/messages`), so a single-protocol relay is not falsely flagged.
- Probes the **real upstream** (unwraps the loopback SSE-sanitize proxy, same as the balance probe).
- Errors classified by status — `auth` (401/403), `model_not_found` (404), `rate_limit` (429), `server` (5xx), `network`/`timeout`; the provider's error message is extracted from the response body and surfaced to the user.
- Status is held in component state (keyed by model id) and cleared on channel/base-URL switch.

## Type of change
- [x] New feature

## Checklist
- [x] Ran `pnpm typecheck` and `pnpm test` (i18n parity 18/18, provider lib tests 31/31)
- [x] Ran `cargo check` + `cargo clippy` + `cargo fmt --check` in `src-tauri`
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW)
- [x] Docs / `docs/llm-wiki` updated (`providers.md` host-command table)
- [x] No secrets included


shiaho777:
image

PR #599, opened 2026-08-13
fix(composer): fork connecting no longer swallows the first send

enderzcx:
## Summary
After **Fork**, warm-connect puts the new chat in handshake `connecting`. That is not a live turn, but Send treated it as busy and parked the first message in the follow-up queue. **Steer** is mid-turn only, so it stayed disabled. The message looked sendable and never left.

- `shouldEnqueueSend` only queues during `streaming` (not `connecting`).
- Queue flush is not held by handshake `connecting`.
- Queue **Steer** becomes **Send now** when no turn is running.

## Test plan
- [x] `vitest` `sendQueue.test.ts` + `sendIntent.test.ts`
- [ ] Fork a finished chat, type a new message, Send → bubble + agent turn (not queue-only)
- [ ] If a leftover queue row exists during connecting, **Send now** dispatches it
- [ ] Mid-turn follow-up still queues; **Steer** still interjects


RongleCat (approved):
Small, correct bugfix. Handshake `connecting` after Fork is not a live turn, so enqueue + disabled Steer swallowed the first send.

- `shouldEnqueueSend` / flush hold now key off a real turn (`streaming`, plus permission for flush).
- Residual queue row becomes **Send now** when Guide is unavailable.
- Tests + en/zh/zh-TW keys look right.

CI reds (frontend 1k budget, Linux `media_server` allowlist) are pre-existing on `main`, not this diff.

Squash-merging. CHANGELOG will be moved to **Unreleased** after land — Git would otherwise place the lines under already-shipped `[0.2.16]`.

RongleCat:
Landed on `main` as da106e5. Thanks @enderzcx — handshake `connecting` after Fork is not a turn, so first Send now waits on `ensureConnected` instead of parking in the queue. CHANGELOG entry moved to Unreleased (would otherwise sit under already-shipped 0.2.16).

PR #603, opened 2026-08-13
fix(terminal): render Starship Powerline glyphs and truecolor

Yy-702:
## Why

The side-workbench PTY already accepted a Nerd Font setting (e.g. `JetBrainsMono NF`), but Starship Catppuccin / Powerline prompts still collapsed to monochrome text plus leftover triangles.

Two WebKit / xterm 6 issues stacked:

1. Unquoted multi-word `font-family` values (`JetBrainsMono Nerd Font`) are parsed as three families. WebKit then matches unpatched **JetBrains Mono**, so Powerline Extra glyphs (`` `` ``) do not fill the cell.
2. `@xterm/xterm` 6 defaults to the **DOM renderer**, which ignores `customGlyphs`. Separators are font-only; without a correctly resolved Mono Nerd Font the rounded caps look like a moon and the color bar disappears.

Inherited `NO_COLOR` from the desktop host could also suppress themed SGR even when `COLORTERM=truecolor` was set.

## What

- Quote every multi-word family and expand known shorthands (`JetBrainsMono NF` → Mono / NFM / NF, Mono first).
- Load `@xterm/addon-webgl` after `open()` so Powerline extras are drawn from cell fg/bg (DOM fallback kept).
- `pty_host` now **removes** `NO_COLOR` and sets `FORCE_COLOR=1` / `CLICOLOR_FORCE=1`.
- Disable coding ligatures on the xterm host; refresh after `document.fonts.ready`.
- Settings hint / preview updated (en / zh / zh-TW).

## Verify

- [x] `pnpm exec vitest run src/lib/cssFontFamily.test.ts src/lib/terminalFontPref.test.ts src/lib/sideTerminalTheme.test.ts src/i18n/messages.test.ts`
- [x] `cargo test --lib pty_host::tests`
- [x] Local `pnpm dev`: open side Terminal, restart the tab, Starship pastel-powerline segments render with filled caps (git / package modules still depend on cwd).


Yy-702:
image
这是修复后的效果

RongleCat (approved):
Reviewed. The stacked root causes look right:

1. Unquoted multi-word `font-family` values are split by WebKit, so `JetBrainsMono Nerd Font` can match unpatched JetBrains Mono.
2. xterm 6 DOM renderer ignores `customGlyphs`; WebGL draws Powerline extras from cell colors.
3. Inherited `NO_COLOR` (even empty) kills Starship/truecolor; the PTY must drop it.

`quoteCssFontFamily` / `expandNerdFontAliases` / Mono-first stack are the right shape. i18n keys match in en / zh / zh-TW. `pty_host` tests pass locally. CI typecheck, fmt, clippy, and macOS/Windows cargo tests are green. The two red jobs (FILES_OVER_1K_BUDGET, Linux `media_server` 404 vs 403) are already failing on `main` and are not from this diff.

Residual, not blocking: WebGL + `allowTransparency` may change the wallpaper veil vs DOM canvas. Context-loss disposes the addon without remounting. Fine for this fix.

Squash-merging.

RongleCat:
已合并，谢谢 @Yy-702。侧栏终端 Starship / Powerline 圆角色块现在会按 Nerd Font Mono + WebGL 画出来，宿主带进来的 `NO_COLOR` 也不会再把主题掐掉。

PR #608, opened 2026-08-13
fix(chat): ignore micro scroll deltas on light upward swipes to prevent jitter

shiaho777:
轻微向上滑动会话底部时，画面会持续抖动但无法真正滚动。

已修复：在虚拟滚动器中添加微增量过滤（<2px），避免不必要的重组和视觉抖动。

测试通过：正常大拖拽仍能滚动，微小触摸不再触发抖动。

（分支已推送到 origin/feat/provider-model-test）

No specific issue.


RongleCat (changes requested):
Reviewed against current main. Do not merge this PR as-is. It mixes three unrelated commits on a stale branch; two are already landed, and the remaining cancelled-turn change is incorrect.

1. feat(providers): per-model connection test — already on main as #594 (96000645). providers_test_model, the Providers panel test button, and the en/zh/zh-TW strings are already in origin/main. Re-shipping it here is why the PR is CONFLICTING.

2. fix(chat): ignore micro scroll deltas (<2px) — superseded. The 2px filter is a band-aid on the virtualizer scroll handler. The actual jitter at the tail is pin-lock follow / near-band re-pin snapping scrollTop back to max after a light swipe. A root-cause fix is already on maintainer main (stop bottom jitter when leaving or landing on the tail): do not treat pin-lock / clamp writes as a user fling; re-engage stick only on the hard bottom, not the 100px near band; commit spacer remasure synchronously.

3. fix(chat): show copy/export/regenerate on cancelled EndOfTurnChip — wrong message. ConversationThread returns early for turn_cancelled / turn_end markers and renders only EndOfTurnChip. That marker's content is turn_cancelled|<reason> (or turn_end|<reason>), not the assistant reply. Copy/export on that row would copy the journal marker string. Regenerate is keyed by regenerableAssistantId === m.id, which is the last assistant id in the user turn — never the chip id — so the regen button would not appear anyway. After stop, Host already sets streaming: false on the assistant and appends a separate chip. The assistant bubble then uses the normal actions path. If those actions are missing on a cancelled assistant bubble, that is the bug to fix — not the chip.

Please close this PR (or strip it down to a single unlanded bugfix with a correct target). One concern per PR; do not reuse feat/provider-model-test for unrelated chat work.

RongleCat:
Thanks for the report @shiaho777.

This PR is not mergeable: it reuses the already-landed #594 branch and stacks two unrelated chat commits on top.

- Per-model connection test: already on main via #594.
- Bottom jitter: root-cause fix is now on main (`b52afba1`) — pin-lock / near-band re-pin, not a 2px filter.
- Cancelled-turn actions: they belong on the finished **assistant** bubble (Host already sets `streaming: false`), not on the `EndOfTurnChip` marker (`turn_cancelled|<reason>`). Copying the chip would copy the journal token.

Closing this one. Please open a focused PR if the assistant toolbar is actually missing after Stop.

PR #609, opened 2026-08-13
fix(chat): show copy/export/regenerate actions for user-stopped (cancelled) turns

shiaho777:
## Summary

When a request is manually stopped by the user (turn cancelled), the bottom action bar (copy, export MD, regenerate, copy link) was sometimes not rendered, so the user could not copy or regenerate the partial reply after stopping.

This PR renders the same bottom actions below the `EndOfTurnChip` for cancelled turns, matching the behavior of completed turns.

## 中文说明

用户手动停止请求后（turn_cancelled），底部操作栏（复制、导出 MD、重新生成、复制链接）有时不显示，导致停止后无法复制或重新生成部分回复。

本 PR 在 `EndOfTurnChip` 下方为被取消的 turn 渲染与正常完成 turn 一致的底部操作。

## Changes

- `src/components/lobe-chat/ConversationThread.tsx`: render copy / export MD / copy-link / regenerate actions for `turn_cancelled` markers when content or a regenerate target exists.

## Verification

- No new tests; logic mirrors the existing completed-turn branch.
- Local typecheck passed.

### DIFF
--- a/src/components/lobe-chat/ConversationThread.tsx
+++ b/src/components/lobe-chat/ConversationThread.tsx
@@ -808,6 +808,72 @@ const TranscriptMessageRow = memo(function TranscriptMessageRow({
   ) {
     return wrap(
       <EndOfTurnChip key={m.id} message={m} locale={locale} />,
+      // Show bottom actions (copy, export MD, regenerate) for cancelled turns so the user can still use them after stopping the request.
+      ( () => {
+        const showCopy = !!m.content.trim();
+        const showRegen = !!onRegenerateAssistant && regenerableAssistantId === m.id;
+        if (!showCopy && !showRegen) return null;
+        const deepLink =
+          sessionId != null
+            ? formatMessageDeepLink(sessionId, m.id)
+            : "";
+        const showCopyLink = !!deepLink;
+        if (!showCopy && !showRegen && !showCopyLink) return null;
+        return (
+          <>
+            {showCopy ? (
+              <>
+                <MessageCopyButton
+                  text={m.content}
+                  copyLabel={tr("message.copy")}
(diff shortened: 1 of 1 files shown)


RongleCat (changes requested):
Do not merge. This is the cancelled-turn hunk from #608, split into its own PR, but it still targets the wrong message and does not attach actions.

1. wrap() only takes the node. The second argument (the IIFE that builds copy/export/regen) is discarded. On current main:

   const wrap = (node) => virtualized ? <div ref=...>{node}</div> : node;

   Extra JS args are ignored. After this patch, cancelled turns still render only EndOfTurnChip.

2. Even if wrap forwarded that node, m here is the turn_cancelled / turn_end marker. m.content is turn_cancelled|<reason> (or turn_end|<reason>), not the assistant reply. Copy/export would copy the journal token. regenerableAssistantId is the last assistant id in the user turn (lastRegenerableAssistantId), never this chip id, so the regen button would not appear.

3. After Stop, Host already sets streaming: false on the assistant and appends a separate chip. The assistant bubble then uses the existing actions= path (copy / export / regen). If that toolbar is missing, the bug is on the assistant row (e.g. still streaming, empty content, or hover-only visibility) — not on the chip.

Please close this PR. If you can reproduce a cancelled assistant bubble with no toolbar, open a new PR that fixes that bubble (and add a test around lastRegenerableAssistantId / the actions guard).

RongleCat:
Closing as the cancelled-chip toolbar is a no-op (wrap ignores the second arg) and would copy the journal marker anyway. See review.

shiaho777:
添麻烦了，刚才用了一下自己中转的Grok 4.6，调用工具有点问题

PR #610, opened 2026-08-14
fix(chat): keep remote image cards after turn-end remount

Yy-702:
## Why

Remote https images in assistant leftovers (e.g. a web-fetch stock chart) render while the answer is still streaming, then flip to **预览失败 — 文件可能已损坏或无法读取** (`broken_blob`) as soon as the turn ends.

The file is fine. Host already wrote a valid JPEG thumb under `cache/image-thumbs/`. What breaks is the remount:

1. `session://journal_reconciled` / ready rehydrates the journal and remounts `ImageUi`.
2. First paint uses the live `https://…` src.
3. `resolveChatImageThumb` hits the in-memory cache and swaps `src` to the loopback thumb in a microtask.
4. WKWebView aborts the in-flight https load → `` → `loadFailed` is locked forever.

## What

- First-paint chat cards from `peekChatImageThumb` when the session cache already has a display URL (no https → loopback swap on remount).
- Ignore abort / stale-src `onError` (`shouldApplyChatImageLoadError`); remount `` with `key={resolvedSrc}`.
- Do not wipe a working https original when thumb resolve returns empty (`nextChatCardDisplaySrc`).

## Verify

- [x] `pnpm exec vitest run src/lib/imageThumbClient.test.ts src/lib/mediaLoadPro.test.ts` (42 passed)
- [x] Local `pnpm dev`: reopen the morning-brief chat that showed the eastmoney chart; after the turn (and after switching away/back) the leftover card stays visible instead of “preview failed”.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 CI Release body = this section only (via `scripts/changelog-for-release.py`; no repeated download/install boilerplate).  
 See `docs/llm-wiki/release.md`.
 
+## [Unreleased]
+
+### Fixed
+- **Chat image cards no longer die after the turn ends**: Leftover remote https thumbs (web-fetch charts, etc.) first-paint from the in-memory thumb cache on journal remount. Swapping `src` mid-load used to abort the original `<img>` and lock `broken_blob` (“preview failed”). Abort / stale-src errors are ignored; a working https original is not wiped when thumb resolve returns empty.
+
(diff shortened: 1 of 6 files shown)


RongleCat:
Merged, thanks @Yy-702 — this is the right remount fix.

Journal rehydrate was first-painting the live https URL, then swapping to the cached loopback thumb and locking `broken_blob` on the WKWebView abort. First-paint from `peekChatImageThumb`, ignoring abort/stale-src, and keeping a working https original when thumb resolve is empty should keep leftover cards (charts, web-fetch images) visible after the turn ends.

Landed on `main` as 46429f3e.

PR #611, opened 2026-08-14
feat(usage): TUI-parity session token spend and heatmap usage column

Yy-702:
## Summary
Adds a TUI `/usage`-style **Usage limit** surface in the App, plus richer local token activity on the account heatmap.

- `/usage` and `/cost` (and the context-chip **Usage limit** action) open a GlassModal with weekly SuperGrok quota and this session’s spend since start or last resume: input (cached), output (reasoning), total, model calls, API time, cost.
- Host now forwards `modelCalls`, `apiDurationMs`, `costIsPartial`, and `usageIsIncomplete` from `turn_completed`. Prompt-result usage is tagged `prompt_result` so it is not double-counted.
- Occupancy ring is unchanged. Custom routes hide official weekly quota.
- Account heatmap recent-session table adds a **Tokens / 用量** column using the same `turn_completed` sum as the heatmap; **Context** stays occupancy.
- Codex-style stats strip above the heatmap: total tokens, peak day, longest chat, current / longest streak. Calendar views: **Day · Week · Cumulative** (running-total color).

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (targeted: sessionSpend, slashCatalog, contextUsage, heatmapStats, heatmapUsagePro, i18n)
- [x] I ran `cargo test` in `src-tauri` (usage parse + call-log usage tokens)
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW)
- [x] No `window.confirm` / `prompt` / `alert` for product dialogs
- [x] Docs / `docs/llm-wiki` updated if behavior changed
- [x] No secrets (`secrets.json`, tokens, `auth.json`) included

PR #612, opened 2026-08-14
fix(usage): keep session spend accurate without emptying the modal

Yy-702:
## Summary
Follow-up to #611 (already merged). Local `/usage` could show **cached > input** (e.g. 144k vs 88k) because per-call / fragment packets were summed with the `turn_completed` snapshot. A first filter then dropped real turns and the modal went empty mid-turn.

- Sum only `turn_completed`.
- Ignore fragments with cache and no `modelCalls` when cache > input (or cache with no input).
- Still accept a real snapshot that omits `modelCalls` if cache ≤ input.
- Clamp per-turn cache to input.
- Persist spend in `sessionStorage` so HMR/reload does not wipe totals.
- While a turn is in flight, show that usage updates when it finishes.

Verified locally: 59,780 input / 57,984 cached / 97% hit.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm test` (sessionSpend + i18n)
- [ ] I ran `cargo test` in `src-tauri` (no Rust change)
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW)
- [x] No secrets included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ See `docs/llm-wiki/release.md`.
 - **Account heatmap stats strip**: Codex-style totals — cumulative tokens, peak day, longest chat, current / longest streak — plus a **Cumulative** calendar view (running-total color).
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


RongleCat:
Merged, thanks @Yy-702 — session spend now stays on `turn_completed` snapshots so the modal no longer inflates cache or goes empty mid-turn.

PR #619, opened 2026-08-16
fix(session): skip set_mode after fork so parent chat can send

enderzcx:
## Why

Forking a live chat (e.g. `a51171b5` from `59f5638a`) succeeded at `_x.ai/session/fork`, then Host walked every agent-mode alias on `session/set_mode` (`agent` / `default` / `code` / `normal` / `Agent`) at **45s each**. After fork the CLI is still hydrating parent context, so those RPCs time out.

Handshake sat in **连接中** for ~4 minutes. Host never logged `session/prompt`. The fork painted **思考中**; switching back to the original chat also showed Connecting and could not send.

## What

- Skip the post-open `set_mode` nudge after a successful fork. Spawn already passed `--permission-mode`; parent mode is already on the forked session.
- If a later `set_mode` hits a transport timeout / closed channel / dead stdout, abort the remaining aliases instead of trying all five.

## Test

- `cargo test --lib set_mode_stops_on_transport_errors` (src-tauri)
- Local 0.2.19 rebuild installed; reproduced from `app.log.2026-08-16` (`session/set_mode` id=5…9, 45s each, then `session_open_ok` with no prompt)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ See `docs/llm-wiki/release.md`.
 **中文 · 变更**
 - **README Gatekeeper 说明**：官方 Release 从 v0.2.19 起已签名并公证。`xattr` 仅留给 fork / 旧的未签名包。
 
+### Fixed
+- **Fork no longer pins the parent chat on 连接中**: After `_x.ai/session/fork` the CLI is still hydrating parent context. Post-open `session/set_mode` used to walk every agent-mode alias (`agent` / `default` / `code` / `normal` / `Agent`) at 45s each — ~4 minutes of handshake, no `session/prompt`, both the fork and the original chat stuck on Connecting / Thinking. Fork skips that nudge (spawn already has `--permission-mode`); any later `set_mode` transport timeout aborts the remaining aliases.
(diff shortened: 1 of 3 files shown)


RongleCat (approved):
Reviewed. Root cause is real: after `_x.ai/session/fork` the CLI is still hydrating, and `session/set_mode` walking `agent` / `default` / `code` / `normal` / `Agent` at 45s each can pin both chats on 连接中.

The two-layer fix is the right shape:
- skip the post-open `set_mode` nudge on the fork path (`--permission-mode` already went out on spawn)
- abort remaining aliases on transport death (timeout / channel closed / stdout closed / write failed), but keep walking on unknown `modeId`

Pushed a one-line `cargo fmt` so the existing `prompt_wait_should_timeout` assert in the same file stops failing `cargo fmt --check`. Will squash-merge when CI is green.

Thanks @enderzcx.

RongleCat:
Merged, thank you @enderzcx — fork no longer burns 5×45s on `session/set_mode` and the rustfmt assert is on `main` so later PRs stop failing `cargo fmt --check`.

Issue #621, opened 2026-08-16
[Feature]: Show all saved accounts + remaining quota in the sidebar user menu


hermes87666:
## Problem / motivation

The sidebar user menu (footer avatar) already shows **the current** official account’s plan and remaining SuperGrok quota. Switching to another saved login still requires **Settings → Account → Switch account**.

Power users keep several xAI / Grok identities on one machine (work / personal / team). Opening Settings every time is slow, and there is no way to **compare remaining quota across accounts** before switching.

**中文 · 动机**  
左下角头像菜单只显示**当前号**的套餐和剩余额度。换号必须进「设置 → 账户 → 切换账号」。一台电脑上有多个 xAI 账号时，没法先看各号还剩多少能量再点一下切换。

## Proposed solution

In the **same user menu** (the card under the avatar):

1. List **every saved official account** (the existing `accounts/` snapshots), not only the active one.
2. For each row, show **honest remaining %** + the small quota bar (same honesty rules as the current card: never invent `0%` / `100%` when the probe failed).
3. Mark the active row as **Active / 当前**.
4. Click a non-active row → `account_switch` (same host path as Settings; recycle agents with `account_auth`).
5. Click the active row → still open **Settings → Account** (today’s behavior).
6. Custom-provider / unsigned-out paths stay as they are (DeepSeek balance, login, etc.).

Suggested host API (small, parallel, best-effort):

- `accounts_quota` → `{ items: [{ id, remainingPercent, usedPercent, resetsAt, available }], fetchedAt }`
- Fetch when the menu opens; seed the current row from the live `account_status` billing snapshot so the active card does not flash empty.

i18n: add keys in **en / zh / zh-TW** (e.g. `account.switchTo`).

**中文 · 方案**  
就在头像弹出的那张卡片里列出全部已保存官方账号；每行显示剩余能量百分比和细进度条；当前号打「当前」；点其他号直接切换（走现有 `account_switch`）；点当前号仍进账户设置。额度探测失败时显示「—」，不要编造 0%。

## Alternatives considered

- Only rename saved-account **labels** to include `"72% remaining"` — visible in the Settings switcher, but stale, mixes identity with quota, and does **not** appear in the user menu.
- Keep switching only in Settings — already works; does not solve “glance + switch from the footer”.
- Fetch quota only for the active account — no comparison, which is the main pain.

## Additional context

[...]

PR #625, opened 2026-08-16
feat(account): list saved accounts and remaining quota in the user menu

hermes87666:
## Summary

Implements [#621](https://github.com/RongleCat/grok-app/issues/621) on current `main`.

The sidebar user menu lists every saved official account with honest SuperGrok remaining %. Click a non-active row to `account_switch`. Click the active row to open **Settings → Account**. The full switcher stays in Settings.

## Constraints (from triage)

- Reuse `accounts_list` / `account_switch` / existing SuperGrok honesty helpers (`accountQuotaHonesty`, `fetch_quota_best_effort`).
- Never invent `0%` / `100%` when the probe failed — show **—**.
- Fetch other accounts when the menu **opens** (best-effort). Seed the active row from the live `account_status` snapshot so it does not flash empty.
- Active row click still opens Settings → Account. Custom-provider / signed-out cards stay as they are.
- i18n: `en` / `zh` / `zh-TW` (`account.switchTo`). No native `<select>` / `window.confirm`.

`accounts_quota` is a small Host helper so inactive snapshots can be probed in parallel without switching. It does not invent percents (`source == error|empty` or `last_error` → unavailable).

## Test plan

- [x] `vitest` for `accountSwitcherQuota` + i18n catalogs
- [x] `tsc --noEmit`
- [ ] `cargo test` `account_profiles` (Host: failed probe hides %; real 0% remaining is kept)
- [ ] Desktop: open user menu with 2+ official snapshots — current row has live % immediately; others fill or stay **—**
- [ ] Click non-active row → switches via existing `account_switch` + agent recycle
- [ ] Click active row → Settings → Account
- [ ] Custom provider / signed-out → original single card (DeepSeek balance / login)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Added
+- **User menu lists every saved official account (#621)**: Each row shows honest SuperGrok remaining % (never invents 0% / 100% on a failed probe). Click another account to `account_switch`; click the active row for Settings → Account. Custom-provider / signed-out cards stay as they are.
+
+**中文 · 新增**
+- **头像菜单列出全部已保存官方账号（#621）**：每行显示诚实的 SuperGrok 剩余 %（探测失败不编造 0% / 100%）。点其他号走 `account_switch`；点当前号仍进设置 → 账户。自定义提供商 / 未登录卡片不变。
+
 ### Changed
 - **README Gatekeeper copy**: Official Releases from v0.2.19 are Developer ID signed and Apple-notarized. The `xattr` workaround stays for forks / older unsigned builds.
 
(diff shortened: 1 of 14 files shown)


RongleCat (approved):
Reviewed against the #621 triage.

- Reuses `accounts_list` / `account_switch` / `fetch_quota_best_effort`
- Probe failure → `available: false` and **—** (does not invent 0% / 100%)
- Active row seeded from live `account_status`; click still opens Settings → Account
- Custom-provider / signed-out stay on the original single card
- i18n en / zh / zh-TW (`account.switchTo`); buttons + existing menu CSS

Rebased onto current `main` and added the Unreleased CHANGELOG note. Host tests `quota_item_*` and the vitest honesty helpers pass locally.

Thanks @hermes87666.

Issue #626, opened 2026-08-16
[Feature] External session-targeted enqueue/wake API for integrations


isCopyman:
## Problem

Grok App can already keep busy turns running after switching chats, maintain a follow-up queue, and run its own scheduled/background work. However, I could not find a public host interface that lets an external local integration deliver a message to one exact existing Grok App session and wake/resume that session when needed.

For example, an external mailbox or agent bus can persist a message addressed to a Grok App session, but the target session does not notice it or start a new turn until the user manually opens the session and sends another prompt. This makes reliable cross-app agent collaboration difficult.

This is specifically about **external, session-targeted enqueue/wake**, not the existing internal background-task wake path.

## Reproduction / current limitation

1. Open a Grok App session and let it become idle or leave it unfocused.
2. Have an external local process persist a message intended for that session (AgentBus is one example).
3. Try to notify Grok App that the exact session has new work.
4. There appears to be no supported local API, CLI, MCP tool, or IPC endpoint for doing so.
5. The message remains outside the Grok App turn lifecycle until the user manually activates the session.

## Expected behavior

A small authenticated host API should accept at least:

- a stable Grok App session identifier;
- message content or a stable content reference;
- an idempotency key;
- a delivery mode such as `enqueue_next_turn` (with any more aggressive mode remaining explicit).

Suggested behavior:

- Persist the request before attempting dispatch.
- If the target session is busy, place it in that session's existing follow-up queue.
- If it is idle, dispatch the next turn.
- If it is unloaded but resumable, resume the **same** session and then dispatch.
- Do not require the Grok App window or chat to be focused.
- Return truthful stages such as `persisted`, `queued`, `runtime_started`, `turn_started`, and `outcome_unknown`.
- Deduplicate retries using the idempotency key.
- Expose lifecycle events/receipts so the sender can distinguish “host accepted” from “model received” and “reply completed”.

The transport could be a local CLI, MCP tool, or authenticated IPC/HTTP endpoint as long as they share the same host-owned implementation.

## Safety considerations

- Local authentication and per-session authorization.
- No silent interrupt/cancel of a running user turn.
- External urgency should affect notification or ordering only unless the user explicitly authorizes interruption.
- A crash after submission must not silently replay the prompt and cause duplicate work.

## Why this is useful

[...]

PR #629, opened 2026-08-16
feat(session): local list + continue-by-id API (#626 first slice)

RongleCat:
## Why

#626 asked for an external way to talk to Grok App sessions. The first slice we agreed on: **list chats**, then **continue the same chat** with `session id + prompt`. Not create. Not interrupt a running turn.

## What

- Host loopback HTTP on `127.0.0.1:0`, token file `{app_data}/session-api.json` (`0600`)
  - `GET /v1/health`
  - `GET /v1/sessions`
  - `POST /v1/sessions/{id}/turns` `{ prompt, idempotencyKey? }`
- CLI (exits **before** Tauri / single-instance so it does not steal focus):
  - `grok-app --sessions`
  - `grok-app --session-send <id> --prompt "…"`
- List can read the on-disk index when the app is quit. Send requires the app or tray.
- Honest statuses: `turn_started` / `busy` / `not_found` / `app_not_running` / `error`
- Settings → Runtime → Connection card (token never shown)
- Wiki: `docs/llm-wiki/session-api.md`

## Out of scope (still #626)

- Creating a new session from outside
- Disk queue while the app is quit
- Interrupt / steal a mid-turn

## Test

- `cargo test --lib session_api` (5)
- `vitest` settings catalog + i18n (35)
- `tsc -b`


RongleCat:
Reviewed against #626 first-slice contract.

- Loopback-only (`127.0.0.1:0`), token-gated including `/v1/health`
- Token file unix `0600`; Settings never prints the token
- Continue-by-id only: no create, no interrupt (`busy` 409)
- Trusted project + `path_ok` required before `connect`
- settingsCatalog + en/zh/zh-TW + wiki

CI green (frontend / Rust mac / linux / win). Squash-merging.

PR #630, opened 2026-08-16
feat(session): queue external prompts while a turn is running

RongleCat:
## Why

After #629, an external `session-send` while the chat was mid-turn (drawing, tools, streaming) returned `busy` and the GUI never saw the prompt.

## What

- If that App session is mid-turn, Host **does not interrupt**. It emits `session://send_queue` and returns `queued` (HTTP 202).
- The main window merges the item into the **same composer follow-up queue** (`source: external`).
- The strip above the composer updates immediately (badge **External** / **外部**). The prompt sends when the turn ends.
- Secondary windows do not listen, so the same item is not sent twice.
- Idempotent retry re-fans the same `itemId`; the GUI dedups.

## Test

- `cargo test --lib session_api` (6)
- `vitest` sendQueue + i18n
- `tsc -b`


RongleCat:
Reviewed: mid-turn goes to the existing composer follow-up queue (`queued` 202), no interrupt, secondary windows do not listen. CI green. Squash-merging.

PR #631, opened 2026-08-16
fix(chat): fold process diaries and recover markdown image previews

enderzcx:
## Why

Long Grok 4.6 turns write first-person progress into `content` (the real thought stream is already on the Worked-for rail). The transcript mashed that diary into the conclusion, so **团队入口** / **01 ·** never read as the answer.

The same turns cite project-relative screenshots (`03 巨字`). Path maps only recognized `images/`-style tails, so MarkdownChat mounted a dead ImageUi on the raw relative src (empty alt card). A first-paint 404 / path-scope miss then locked the card, so the preview looked lost even after the file existed.

## What

- Split the assistant body at the first process-like heading, standalone `---`, or `**01 ·`. Fold the prefix into a muted **过程 / Process** row (same chrome as progress notes). Visible markdown starts at the conclusion.
- Map attachment tails / basenames so `design-demos/shots/foo.png` resolves. Do not mount ImageUi on an unresolved relative href.
- Retry local preview resolve/load a few times (file still being written, grant not ready) instead of locking broken on the first miss.

## Test

- `pnpm exec vitest run src/lib/assistantAnswerSplit.test.ts src/lib/attachments.test.ts src/lib/mediaLoadPro.test.ts src/i18n/messages.test.ts`
- Local 0.2.19 rebuild installed; check the Viko fork session: diary folded, **03 巨字** paints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,14 @@ See `docs/llm-wiki/release.md`.
 
 ### Changed
 - **README Gatekeeper copy**: Official Releases from v0.2.19 are Developer ID signed and Apple-notarized. The `xattr` workaround stays for forks / older unsigned builds.
+- **Assistant process vs answer**: First-person work diaries inside a single reply fold into a muted Process row (same chrome as progress notes). The visible markdown starts at the first heading, `---`, or `**01 ·` after that diary, so the conclusion is no longer mashed into the work log.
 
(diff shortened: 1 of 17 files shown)


RongleCat:
Reviewed and squash-merged. Thanks @enderzcx — process/answer split plus the relative-image retry path are in.

PR #633, opened 2026-08-16
feat(slash): add /workflow and /workflows composer commands

sk1935:
## Summary
Grok Build CLI already has `/workflow` and `/workflows`. The desktop App slash palette did not list them, so they looked missing.

This adds both commands to the composer `/` menu:
- `/workflows` or bare `/workflow` → Settings → Runtime → workflows card (saved scripts + smoke/run). The App still has no TUI run dashboard.
- `/workflow <name|pause|resume|stop|save …>` → sent as a session turn so the agent / CLI host can run it.

Typing a lone `/workflows` / `/workflow` and pressing Enter also opens the settings card instead of sending a dead prompt.

## Type of change
- [x] New feature

## Checklist
- [x] I ran `pnpm typecheck` and targeted `pnpm exec vitest` (workflowSlash / slashCatalog / draftDoc / i18n) — 96 passed; `tsc` clean
- [ ] I ran `cargo test` in `src-tauri` (frontend-only; no Rust on this machine)
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW)
- [x] No `window.confirm` / `prompt` / `alert` for product dialogs
- [x] CHANGELOG Unreleased updated
- [x] No secrets included

Full `pnpm test` had 6 unrelated failures here (canvas hook timeout, desktopNotify quiet-hours timezone, remoteIm mock). Slash-related files passed.

Issue #618, opened 2026-08-16
[Bug]: 第三方 Grok API 无法读取图片附件


jackBaozz:
## 问题描述

使用第三方 Grok API（自定义 API / relay）时，在 Grok App 会话中上传或粘贴图片，模型无法读取图片内容。

## 复现步骤

1. 在 Grok App 中配置第三方 Grok API 或自定义中继服务。
2. 选择支持视觉输入的 Grok 模型。
3. 在会话中上传或粘贴 PNG/JPG/WebP 图片。
4. 发送消息并要求模型识别图片内容。

## 实际结果

图片没有被模型正常识别，并出现类似以下错误：

```text
Hook denied: read_file on image files is blocked on custom text-only main: tool results embed image_url and the relay API returns 400 (unknown variant image_url). Report the absolute path as plain text for the user/UI. Vision is already provided via Host vision or official-aux__vision_describe — do not re-open generated PNG/JPG/WebP.
```

其中关键错误是 relay API 返回：

```text
400 (unknown variant image_url)
```

## 期望结果

当第三方 API 和所选模型支持视觉输入时，应用应将图片按照该 API 兼容的格式传递给模型，例如标准的多模态 content 结构、data URL/base64 或该中继服务要求的格式。

如果第三方 API 或模型不支持图片输入，也应该在发送前给出明确提示，而不是只返回难以理解的 `unknown variant image_url`。

## 建议

- 按 provider / model 能力判断是否支持 image input / vision。
- 为第三方 OpenAI-compatible relay 提供图片请求格式适配或可配置选项。
- 对 `image_url`、base64 和 multipart 等常见格式进行兼容处理。
- 在 API 返回 400 时展示更具体的原因和排查建议。

## 环境信息

- 系统：macOS
- 应用：Grok App
- API：第三方 Grok API / 自定义 relay
- 模型：待补充
- 图片类型：PNG

RongleCat:
## Triage

**P1 · `area:session`** — the pasted text is our **PreToolUse hook deny reason**, not necessarily a live relay 400.

On a **custom / third-party main**, Grok App treats unknown providers as **text-only** so the CLI does not inject `image_url` into DeepSeek-style `chat_completions` (that really is `400 unknown variant image_url`). Images are stripped to path notes; Host vision / `official-aux__vision_describe` is supposed to describe them. The hook then **blocks `read_file` on `.png/.jpg/.webp`** so the agent cannot re-open the file and blow up the next request.

[...]

PR #617, opened 2026-08-15
fix: preserve Grok Build relay semantics through CC Switch

enderzcx:
## What changed

- add an explicit `grok_build_proxy` provider mode for relays that implement Grok Build's native model catalog and chat proxy contract
- launch ACP with the real remote model id and process-scoped native catalog/proxy environment, preserving backend Web/X search semantics
- validate native mode against the live `/models` response and require `supports_backend_search=true` for every selected model
- auto-promote CC Switch `grokbuild` imports only when the selected Responses model advertises that capability
- keep ordinary OpenAI/DeepSeek-compatible providers on the existing generic `[model.*]` path
- recycle warm ACP children after provider edits/imports so stale route semantics cannot be reused

## Review feedback addressed

1. Rebased onto current `main` (`5e0a46e`, including #619 / #622 / #624 / #625 / #627 / #628 / #629 / #630 / #632).
2. Added a direct opt-in path: **Settings → Account → Custom providers → Provider mode → Grok Build-compatible relay**. The setting has its own Settings search entry. After a live model fetch, selected models missing `supports_backend_search=true` are shown inline; save also fails closed against a fresh backend probe.
3. Attachment scope is explicit: this mode does not add or claim native ACP image content blocks. Grok App continues to send attachments through the existing `@absolute/path` text flow from #627.
4. Removed the installed-binary hash and machine-specific install detail from this PR description.

## Root cause

Grok App wrote every custom relay as a generic `[model.<alias>]` provider and spawned ACP with that alias. This discarded Grok Build's native catalog/proxy semantics even when the relay advertised backend search, so the CLI could not attach its backend Web/X tools. CC Switch imports reinforced the issue by hard-coding every imported provider to generic mode.

## Security and compatibility

- no provider-name or hostname special cases
- provider keys remain write-only and are injected only into the target Grok child process
- no official xAI credential or `official-aux` MCP fallback is required
- generic providers retain their existing behavior
- WSL forwards native proxy variables without path translation
- native mode is Responses-only and fails closed when the live catalog cannot prove capability

## Validation

- `pnpm test`: 386 files / 5526 tests passed
- `pnpm typecheck`: passed
- `pnpm lint`: passed
- `python3 scripts/check-code-quality-gates.py --mode final`: passed
- `pnpm build:ui`: passed
- `cargo fmt --all -- --check`: passed
- `cargo clippy --all-targets -- -W clippy::all`: passed (existing warnings only)
- `cargo test`: 1217 passed, 1 ignored
[...]

PR #634, opened 2026-08-16
fix: preserve Grok Build relay semantics through CC Switch

enderzcx:
## What changed

Supersedes #617 after the maintainer-requested rebase and review fixes. Related: #618.

- add an explicit `grok_build_proxy` provider mode for relays that implement Grok Build's native model catalog and chat proxy contract
- launch ACP with the real remote model id and process-scoped native catalog/proxy environment, preserving backend Web/X search semantics
- validate native mode against the live `/models` response and require `supports_backend_search=true` for every selected model
- auto-promote CC Switch `grokbuild` imports only when the selected Responses model advertises that capability
- keep ordinary OpenAI/DeepSeek-compatible providers on the existing generic `[model.*]` path
- recycle warm ACP children after provider edits/imports so stale route semantics cannot be reused

## Review feedback addressed

1. Rebased onto current `main` (`719c72b`, including #619 / #622 / #624 / #625 / #627 / #628 / #629 / #630 / #631 / #632 / #633).
2. Added a direct opt-in path: **Settings → Account → Custom providers → Provider mode → Grok Build-compatible relay**. The setting has its own Settings search entry. After a live model fetch, selected models missing `supports_backend_search=true` are shown inline; save also fails closed against a fresh backend probe.
3. Attachment scope is explicit: this mode does not add or claim native ACP image content blocks. Grok App continues to send attachments through the existing `@absolute/path` text flow from #627.
4. Removed the installed-binary hash and machine-specific install detail from this PR description.

## Root cause

Grok App wrote every custom relay as a generic `[model.<alias>]` provider and spawned ACP with that alias. This discarded Grok Build's native catalog/proxy semantics even when the relay advertised backend search, so the CLI could not attach its backend Web/X tools. CC Switch imports reinforced the issue by hard-coding every imported provider to generic mode.

## Security and compatibility

- no provider-name or hostname special cases
- provider keys remain write-only and are injected only into the target Grok child process
- no official xAI credential or `official-aux` MCP fallback is required
- generic providers retain their existing behavior
- WSL forwards native proxy variables without path translation
- native mode is Responses-only and fails closed when the live catalog cannot prove capability

## Validation

- `pnpm test`: 386 files / 5526 tests passed
- `pnpm typecheck`: passed
- `pnpm lint`: passed
- `python3 scripts/check-code-quality-gates.py --mode final`: passed
- `pnpm build:ui`: passed
- `cargo fmt --all -- --check`: passed
- `cargo clippy --all-targets -- -W clippy::all`: passed (existing warnings only)
- `cargo test`: 1217 passed, 1 ignored
- targeted provider / CC Switch / ACP env tests: 23 + 13 + 1 passed
- local Settings UI inspection: explicit mode selector, Responses coercion, honest capability/attachment copy, and solid Select menu verified

One initial full Rust run hit three timing-sensitive failures outside this provider path; all three passed when replayed individually, and the complete suite passed on the next run.

## Live acceptance boundary

The same route implementation previously passed isolated Windows BeefAPI acceptance: `/models` advertised backend search for Grok 4.5/4.6, Web search emitted native tool lifecycle events, and X search returned a live `x.com/status/...` result. That credential was not persisted or printed. The post-rebase review pass used deterministic tests and did not reuse a production token.

## ACP boundary

ACP remains the recommended primary GUI interface because it preserves Grok Build's session, tool, permission, MCP, and streaming semantics without reimplementing the agent brain. The audit and remaining protocol/live-evidence gaps are documented in `docs/SPIKE-ACP.md`.


RongleCat:
Reviewed and squash-merged. Thanks @enderzcx — Settings opt-in, live `supports_backend_search` gate, and process-scoped native catalog env are in. CHANGELOG Unreleased updated on main.

PR #635, opened 2026-08-16
fix(chat): fold process speech and actions inside 工作了

enderzcx:
## Why

#631 folded first-person diaries into a separate **过程** strip. The conclusion still sat under a tools-only **工作了** rail, so a long Grok 4.6 turn looked like: thought titles + tools, then another process blob, then the answer. Grok Build’s stream is **speech → folded actions → speech → folded actions**, with the conclusion always visible.

## What

Follow-up to #631 (image-preview fixes stay on `main`).

- Mid-turn `content` and the process half of the last body go **inside** one **工作了** fold (default collapsed).
- Inside the fold: user-facing speech interleaved with collapsed action groups (探索 / 运行了 N 个命令 / 编辑了 N 个文件). Individual rows still expand.
- Journal CoT is hoisted to a separate **思考了** row — not dumped into the rail.
- The last-content conclusion stays **below** the fold.

## Test

- `pnpm exec vitest run src/lib/timelinePhases.test.ts src/lib/assistantAnswerSplit.test.ts src/lib/grokActivitySteps.test.ts src/i18n/messages.test.ts`
- `pnpm typecheck`
- Local 0.2.19 rebuild installed; open a long Grok 4.6 turn: collapsed `工作了` + conclusion, expand to speech ↔ folded actions.

## 中文

过程（中间说的话 + 折叠动作）收进同一条「工作了」，默认收起；结论一直在下面。「思考了」仍单独一行。

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,11 +29,11 @@ See `docs/llm-wiki/release.md`.
 
 ### Changed
 - **README Gatekeeper copy**: Official Releases from v0.2.19 are Developer ID signed and Apple-notarized. The `xattr` workaround stays for forks / older unsigned builds.
-- **Assistant process vs answer**: First-person work diaries inside a single reply fold into a muted Process row (same chrome as progress notes). The visible markdown starts at the first heading, `---`, or `**01 ·` after that diary, so the conclusion is no longer mashed into the work log.
+- **Assistant process vs answer**: Mid-turn speech and folded actions (explore / run / edit) live inside one **工作了** fold (collapsed by default). The conclusion stays visible below. Journal CoT stays a separate **思考了** row — not dumped into the rail or mashed into the answer.
 
 **中文 · 变更**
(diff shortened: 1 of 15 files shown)


RongleCat:
Reviewed with pi (read-only). Direction is right — #631’s extra **过程** strip on top of **工作了** was the wrong chrome.

Two merge blockers are fixed on this head before squash:

1. **Last content after tools** — `hasWorkAfter` no longer swallows the whole conclusion into 工作了. The last content is always the answer slot; trailing tools (open Finder, screenshot) stay in the work fold.
2. **Find** — Cmd+F occurrence indexes skip the folded process prefix of `m.content`, so conclusion highlights line up. Speech inside 工作了 also highlights. Blank-line cuts now need a first-person / in-progress marker so a long “已经/接下来是结论 + 表” is not hidden.

Squash-merged. Thanks @enderzcx.

PR #652, opened 2026-08-17
fix: harden session interruption recovery

dundunge:
## Summary

- Harden ACP process shutdown and TCP reader cancellation so recycled agents do not leave a blocked reader or emit ghost events.
- Make stamped ACP events require an exact agent-session owner, with regression coverage for unbound live/background hints.
- Isolate per-session send claims and epochs, preserve the viewed session during mirror rehydration, and clean draft-send state transitions.
- Retry external `session://send_queue` listener registration after transient IPC failures.
- Preserve the existing journal, mirror, relay, and process-lifecycle recovery fixes from the audit implementation.

## Root cause

The interruption path mixed process-scoped state with session-scoped state. A kill could race the TCP reader before `Notify` registration, while late stamped events could be accepted when the current slot had no recorded owner. Frontend hydration and queue listeners also assumed a single successful registration, allowing stale focus or permanently missed follow-ups after reconnects.

## Validation

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test -- --testTimeout=15000` 鈥?389 files, 5571 tests passed
- `pnpm vitest run src/lib/sendQueue.test.ts --testTimeout=15000` 鈥?30 tests passed
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- Rust compile/test execution is environment-blocked here: MSVC `link.exe` and GNU `dlltool.exe` are not installed.

### DIFF
--- a/src-tauri/src/acp_client.rs
+++ b/src-tauri/src/acp_client.rs
@@ -14,7 +14,7 @@ use serde_json::{json, Value};
 use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader};
 use tokio::net::TcpStream;
 use tokio::process::{Child, Command};
-use tokio::sync::{mpsc, oneshot, Mutex as AsyncMutex};
+use tokio::sync::{mpsc, oneshot, Mutex as AsyncMutex, Notify};
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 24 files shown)


RongleCat:
Review pass after rebase onto current main. Kept quote payload on queue flush; send retry uses journalDisplay. Landed on main.

PR #655, opened 2026-08-17
feat(workbench): Design Mode, agent Kanban, and project Spaces

Yy-702:
## Why

The workbench needed three related surfaces: inspect a local preview without leaving the Browser tab, see running agents by status, and group sidebar projects without losing that grouping on restart.

## What

- **Design Mode** — toolbar toggle on the Browser tab. Hover/click a same-origin or localhost element, then send tag, CSS path, computed styles, optional snapshot, and a style note into the composer.
  - `side_browser_eval` / `snapshot` are async so a navigating WebView does not freeze the app.
  - iframe / `pageLoading` flicker no longer tears down a live overlay or clears the selection.
  - In-flight install still tears down on disable/unmount.
  - Large captures that fall back to JPEG keep `image/jpeg` / `.jpg`.
- **Agent Kanban** — sidebar / palette / `#/kanban`. Columns are Needs you / Working / Done (Idle hidden by default). Cards are live sessions, not personal to-dos.
  - A finished turn stamps `terminalReason` and is remembered so Done survives leaving the pane.
  - Opening a card does not hide it; a new turn moves it back to Working.
  - Card click resolves the session (including a list refresh) before leaving the board.
- **Project Spaces** — All / Default / named spaces. Membership is stored in App settings.
  - Space writes are coalesced so “move to new space” is one snapshot (no lost membership after restart).
  - Edits before settings hydrate are kept.
  - Search from All does not switch space.
  - Name errors (duplicate, empty, too long) stay inline in the prompt in `--danger` red, not a toast. (`.modal p` no longer overrides the error color.)

## Verify

- `npx vitest run src/lib/kanbanBoard.test.ts src/lib/sessionFinishedTurns.test.ts src/lib/sessionLiveStore.test.ts src/lib/projectSpaces.test.ts src/lib/browserDesignMode.test.ts src/lib/paletteActions.test.ts`
- `npx tsc -b`
- Manual: Design Mode on a localhost preview (select → send; iframe flicker keeps selection; toggle off restores clicks). Kanban: finish a turn → see Done → leave and come back. Spaces: create / rename / move / restart; duplicate name shows red inline next to Cancel / Save.

## Notes

i18n: en / zh / zh-TW. Dialogs use `setAppDialog` (no `window.prompt`). CHANGELOG Unreleased updated.

PR #657, opened 2026-08-17
fix(search): restore palette wheel scroll and add keyboard selection

Yy-702:
## Why

Command-palette search (`⌘K`) could not be scrolled with a mouse wheel or trackpad. Hits sat in `.search-panel` under `max-height` + `overflow: hidden` with no inner scrollport, so the list was clipped and wheel/trackpad had nowhere to go (same trap as automations #543). After that, the list also had no keyboard highlight.

## What

- Wrap actions / projects / sessions in `OverlayScroll.search-panel__results` so the search field and filter chips stay put.
- Pin chrome with `flex-shrink: 0`; results viewport is the only `overflow-y: auto` scroller.
- `↑` / `↓` move a highlight across the flattened list (wraps); `Enter` opens the active row; `PageUp` / `PageDown` jump 8 rows; `⌘1`–`⌘9` (`Ctrl` on Windows) open the Nth **session** (same numbering already shown in the row meta).
- `Home` / `End` stay with the query caret. Changing the query or filters resets the highlight; late content hits only clamp the index.
- Mouse hover updates the highlight. Active row uses the existing menu hover surface and `scrollIntoView({ block: "nearest" })`.

## Verify

- `pnpm exec vitest run src/lib/searchPanelNav.test.ts src/lib/searchPanelScroll.guard.test.ts`
- Manual: open search, type a query that overflows the panel, wheel / trackpad the list. Then `↓` / `↑` and `Enter`. Confirm `⌘1` opens the first session row.

## Notes

Independent of #655 (Design Mode / Kanban / Spaces). No new i18n strings. No `window.confirm` / native `<select>`.

### DIFF
--- a/src/app/AppWorkbench.tsx
+++ b/src/app/AppWorkbench.tsx
@@ -428,6 +428,11 @@ import {
   type PaletteActionDef
 } from "@/lib/paletteActions";
 import {
+  flattenSearchPanelItems,
+  type SearchPanelItem,
+} from "@/lib/searchPanelNav";
+import { useSearchPanelNav } from "@/hooks/useSearchPanelNav";
+import {
   canOfferContinueCwd,
   classifyContinueCwdEmptyResult,
   continueCwdSoftFailMessageKey,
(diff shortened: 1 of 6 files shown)


RongleCat:
**Review pass.** No Spaces dependency. Scrollport + keyboard policy are tested. Landed on main after #663 (AppWorkbench auto-merged). Thanks.

PR #672, opened 2026-08-17
fix: stop capped Working lists from crushing rows into overlap

oykb58246:
Follow-up to #668 (already merged).

## Motivation

#668 stopped leftover `[39m` and clipped virtual 30px rows. After merge, a long **工作了** list still looked crushed: expanding one file/tool painted its body over later rows (`探索` / `Use tool`).

## Cause

Leaving `VirtualList` still put speech / expanded steps in a `12 × 30px` flex box (`max-height: 360px`). Default `flex-shrink` plus `min-height: 0` mashed every row into that box, so expand bodies overlapped later tools. The rail was also capped at `48rem`, so labels and file tails felt tight.

## Change

- Do not shrink `.grok-act__step` rows
- Mapped (speech / expanded) lists use `min(70vh, 40rem)` instead of `N × 30px`
- Widen the activity rail so labels and file tails have room
- `stripAnsi` skips the regex chain when the dump has no ESC/SGR (Copilot review on #668)

## Verify

- `pnpm exec vitest run src/lib/ansi.test.ts src/lib/grokActivityVirtualize.test.ts`
- Manual: expand **工作了** with 20+ steps → open one file/tool → later rows move down, no overlap

## Env

Windows 11 · Grok App 0.2.20 + local patch · 2560×1600 @ 150% · light theme

### DIFF
--- a/src/components/lobe-chat/TimelinePhaseBlock.tsx
+++ b/src/components/lobe-chat/TimelinePhaseBlock.tsx
@@ -45,11 +45,11 @@ import {
 } from "@/lib/toolDisplay";
 import {
   GROK_ACTIVITY_STEP_ROW_PX,
-  GROK_ACTIVITY_VIRTUALIZE_THRESHOLD,
   applyActivityStepExpandPolicy,
   applyActivityStepUserToggle,
   emptyActivityStepExpandState,
   grokActivityVirtualMaxHeightPx,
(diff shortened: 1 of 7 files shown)


RongleCat:
Review pass. Follow-up to #668: drop the N×30px flex-shrink cap that crushed expanded rows. CI green.

PR #682, opened 2026-08-18
fix: preserve reasoning effort across session respawns

MaxxxDong:
## Summary

Fix reasoning-effort changes that appear in the UI but are overwritten when an existing CLI session is resumed.

- serialize effort preference updates with the next send
- expose native reasoning-effort capability for custom provider effort catalogs
- invalidate the stored CLI agent session only when the effort actually changes
- force `session/new` so fresh `--model` and `--reasoning-effort` flags remain authoritative
- preserve conversation continuity through the existing local journal bootstrap
- improve safe spawn diagnostics and the custom-provider credential error message

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Root cause

Changing the effort soft-respawns the CLI with new process flags. For an existing conversation, the App then resumed the previous CLI agent session with `session/load`.

Grok Build restores the prior model and reasoning effort from that CLI journal, overriding the new process flags. The App therefore displayed the newly selected effort while requests could continue using the previous value.

With some custom API configurations, the restored model identifier can also resolve to a built-in route instead of the intended custom endpoint. If that route has no valid first-party credential context, the App reports an unavailable credential and the ACP connection drops.

This is not provider-specific: it depends on session resume state, model identifier resolution, and whether the custom backend exposes reasoning-effort support.

## Reproduction

1. Configure a custom model using the Responses backend and an effort catalog that includes `high` / `xhigh`.
2. Start a conversation with that custom model and complete one turn.
3. After the turn is idle, change the reasoning effort.
4. Send another message in the same conversation; repeating the switch makes the failure easier to observe.
5. Inspect the custom API trace or CLI session summary.

Before this change:

- the UI shows the new effort
- the resumed CLI session may keep the previous effort
- some custom configurations may restore a different route and surface a credential/disconnection error

Expected:

- the selected effort applies to the next turn
- the custom route remains selected
- the conversation remains usable

## Implementation

On a real effort change, clear the stored `agent_session_id` before the soft respawn. The next connection uses `session/new`, then the existing App journal bootstrap supplies recent conversation context.

The frontend also queues the preference write and awaits it before a direct send, closing the race where a message could reuse the old ready process before the host finished applying the effort.

### Automated

- `pnpm test`: 407 test files, 5,769 tests passed
- `pnpm typecheck`: passed
- `pnpm lint`: passed
- focused Rust regression tests cover custom-provider reasoning capability serialization, migration of existing configs, and invalidation of only the selected CLI session
- `cargo test`: 1,250 passed, 1 ignored; one unrelated existing `plan_chrome::tests::upsert_and_mark_stale_round_trip` test is flaky under the full parallel run and passes when rerun in isolation
- `cargo fmt --all -- --check` currently reports two pre-existing formatting differences in `session_manager/connect.rs` on upstream `main`; this PR does not modify that file

### Manual end-to-end

Tested a custom Responses API in one existing conversation while switching:

`medium → xhigh → high`

For both effort changes:

- the CLI process spawned with the custom model alias and requested effort
- ACP used `session/new`, not `session/load`
- the new CLI session summary recorded the custom model alias and exact selected effort
- inference completed successfully
- no route fallback, credential error, or ACP disconnect occurred

## Checklist

- [x] I ran `cargo test` in `src-tauri`
- [ ] User-facing strings go through `src/i18n/messages.ts` (en + zh)

### DIFF
--- a/src-tauri/src/acp_client.rs
+++ b/src-tauri/src/acp_client.rs
@@ -1571,9 +1571,11 @@ impl AcpClient {
             crate::wsl_backend::apply_wslenv(&mut cmd);
         }
         tracing::info!(
-            "acp: spawn home={} mode={} native_grok_proxy={} sandbox={:?} max_turns={:?} fork_session={} no_ask_user={} leader={} subagents={} memory={} compaction_mode={} compaction_detail={} compaction_flags={} agent_profile={:?} agents_json={}",
+            "acp: spawn home={} mode={} model={} effort={:?} native_grok_proxy={} sandbox={:?} max_turns={:?} fork_session={} no_ask_user={} leader={} subagents={} memory={} compaction_mode={} compaction_detail={} compaction_flags={} agent_profile={:?} agents_json={}",
             grok_home.display(),
             session_data_mode,
+            spawn_model,
+            opts.effort,
             grok_build_proxy.is_some(),
             sandbox.as_ref().map(|s| s.profile.as_str()),
             max_turns.as_ref().map(|m| m.turns),
--- a/src-tauri/src/commands/settings.rs
+++ b/src-tauri/src/commands/settings.rs
@@ -354,6 +354,9 @@ pub async fn composer_prefs_set(
     } else {
         session_id.or(live_sess)
     };
+    let previous_effort = effort.as_ref().map(|_| {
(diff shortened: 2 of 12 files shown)


RongleCat:
谢谢，已进 main。随下一版发。

PR #681, opened 2026-08-18
feat(workbench): add chat-column bottom terminal and pane split motion

Yy-702:
## Summary

Adds a VS Code-style terminal under the chat column and makes sidebar / aside / bottom-pane click-toggle interpolate instead of hard-cutting. Settings, account/env menus, and project session lists now use the same motion language.

- **Bottom terminal:** `⌘\`` now toggles a persist-mounted panel in the chat column (tabs are per-project; open/height are layout). Closing the panel keeps PTY sessions at height 0. The side-workbench picker can still open a terminal tab, but no longer owns the shortcut.
- **Pane motion:** click-toggle writes the same inline size tuple as drag (`width`/`min`/`max`/`flex-basis`, or height for the bottom pane). Desktop CSS no longer forces `width: 0 !important`, which was killing WKWebView transitions. `--motion-pane` is **not** zeroed under `prefers-reduced-motion`.
- **Settle isolation:** native browser webviews, xterm `SIGWINCH`, chat virtual-list, stick-to-bottom, overlay scroll, heatmap, and Office preview wait until the motion window ends so they do not resize every animation frame or paint a stale overflow rect over chat.
- **Settings overlay:** opening Settings keeps the workbench mounted (`inert` + faded) and rises the page from the account-menu corner; closing sinks it back instead of swapping views. Native webviews stay covered while the stage is mounted.
- **Account / env menus:** user-menu pop, theme flyout, and env-info pop stay mounted through exit (`useOpenPresence`) so they interpolate instead of hard-cutting.
- **Project chats:** expanding a project (or Other sessions) interpolates list height (`0fr` → `1fr`) with a short fade/slide instead of mounting the rows as a hard cut.

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (targeted: `bottomTerminal`, `paneSplitMotion`, `nativeWebviewBounds`, `sideWorkbench`, `openPresence`, `floatingMenu`, `SidebarTreeReveal`; plus `pnpm typecheck`)
- [ ] I ran `cargo test` in `src-tauri` (or `cargo check` for docs-only) — no Rust changes
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW)
- [x] No `window.confirm` / `prompt` / `alert` for product dialogs
[...]

PR #694, opened 2026-08-18
fix(sidebar): interpolate project collapse and hide leftover scrollbar

Yy-702:
## Summary

Follow-up to #681. Expanding a project already interpolated; collapsing it (and the L1 Projects chevron) still snapped in WKWebView, and the overlay scrollbar stayed after the list hit height 0.

- **Close path:** lock the used px height, paint it, then interpolate to 0. Writing `auto→0` in one commit is the same snap as `0→N` before paint.
- **Stay on the inline px tuple** through settle so the next close is not another `auto→0` snap.
- **L1 Projects section:** wrap `projectsOpen` in `SidebarTreeReveal` so the whole project list interpolates, not only a single project's chats. Session-row motion stays on `--motion-normal` (200ms).
- **Scrollbar:** while the list is interpolating, hide the overlay thumb, force `overflow: hidden`, and defer OverlayScroll measure / VirtualList scroll-into-view. Remeasure on settle so the rail does not flash or linger.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (targeted: `treeReveal`, `SidebarTreeReveal`; plus `pnpm typecheck`)
- [ ] I ran `cargo test` in `src-tauri` (or `cargo check` for docs-only) — no Rust changes
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh) — no new strings
- [x] Docs / `docs/llm-wiki` updated if behavior changed (`docs/design-tokens.md` notes sidebar project-list motion)

## Test plan
- [ ] Click a project row: list height interpolates open; collapse plays the reverse without snapping
- [ ] Click the L1 Projects chevron: the whole project list interpolates; siblings below slide instead of jumping
- [ ] After collapse, no leftover overlay scrollbar / WKWebView indicator on the rail
- [ ] Other sessions uses the same expand/collapse motion
- [ ] Rapid expand/collapse does not flash the rail thumb or jump the virtual list

### DIFF
--- a/docs/design-tokens.md
+++ b/docs/design-tokens.md
@@ -37,8 +37,8 @@
 | `--text-xs` / `sm` / `md` / `lg` | `12 / 13 / 14 / 16 px` | 字号 |
 | `--leading-tight` / `normal` | `1.35 / 1.55` | 行高 |
 | `--shadow-window` | 见主题表 | 窗口投影 |
-| `--motion-fast` / `normal` | `120ms` / `200ms` | 过渡（含项目会话列表开合）；尊重 `prefers-reduced-motion` |
-| `--motion-pane` / `--motion-pane-ease` | `320ms` / `cubic-bezier(0.22, 1, 0.36, 1)` | 左右分栏 / 底栏开合 / 设置页叠在工作台上的进出（无过冲）。分栏 token **不**随 `prefers-reduced-motion` 清零；设置 overlay 与菜单在 reduce 下关掉 transform / transition |
+| `--motion-fast` / `normal` | `120ms` / `200ms` | 过渡；尊重 `prefers-reduced-motion` |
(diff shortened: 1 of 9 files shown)


RongleCat (approved):
## Review

Follow-up to #681 looks correct: WKWebView only interpolates concrete px, so close must paint a locked height then go to 0 (same class of snap as `auto→0` / `0fr/1fr`). Staying on the inline px tuple after settle, wrapping the L1 Projects list, hiding the overlay thumb for the motion window, and deferring OverlayScroll / VirtualList work are the right levers.

Frontend CI is green. Rust jobs fail on `cargo fmt` in `pet_window.rs`, which this PR does not touch (pre-existing on `main`; already fixed locally and will land separately).

**Nits we will handle on `main` after squash:**
- Add an Unreleased CHANGELOG entry.
- Rebase onto unpublished session-move work that also touches `SidebarTreeReveal` (grow-on-content when a chat is dropped into an open folder). Prefer 694’s stay-on-px ResizeObserver over releasing to `auto`.

Squash-merging.

RongleCat:
Merged, thank you @Yy-702 — collapse now interpolates and the leftover overlay scrollbar is gone. CHANGELOG Unreleased will pick this up on `main`.

PR #699, opened 2026-08-19
fix(session): unlock composer when a finished turn stays on 思考中 / 连接中

oykb58246:
## Summary
After the last assistant reply is already on screen, the workbench can stay on **思考中** (quiet thinking) and **Stop**, with the header stuck on **连接中**. The next send never leaves because `session.state` is still `streaming` and/or a leftover `ensureConnected` claim is held.

Ghost heal only strips *empty* optimistic shells, so a finished body never recovered.

This PR unlocks the composer **without deleting the reply** when:
- Host/liveMap is already idle/ready (8s grace), or
- the transcript is done but Host itself stayed streaming/connecting for 90s
- a 连接中 claim exists with no handshake actually running

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] Targeted `pnpm test` (`zombieBusyHeal`, `ghostStreamingHeal`, i18n catalog)
- [ ] Full `cargo test` (no Rust change)
- [x] User-facing strings via i18n (en + zh + zh-TW; ru override)
- [x] CHANGELOG Unreleased updated
- [x] No secrets

## Test plan
- [ ] Send a turn, wait until the full reply is painted; if 思考中 + Stop linger, they should clear within ~8s and Send should work
- [ ] A live think→tool loop (assistant still `streaming`) must **not** be healed
- [ ] Header 连接中 with no in-flight handshake should drop so the next send is not blocked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Fixed
+- **Finished turns no longer stay on 思考中 / 连接中**: After the last assistant body is painted, leftover streaming or a leaked connect claim kept Stop and blocked the next send. Client heal unlocks the composer (reply stays) when Host is idle, or after 90s if Host itself went stale.
+
+**中文 · 修复**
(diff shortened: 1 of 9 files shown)


RongleCat:
Merged — thanks @oykb58246. Finished turns that kept 思考中 / 连接中 will now unlock the composer without deleting the reply.

PR #708, opened 2026-08-19
feat(i18n): expand to 17 locales and remove the zh/en assumptions

1llum1n4t1s:
## Motivation

Grok App ships four languages today (`en`, `zh`, `zh-TW`, `ru`). This PR takes it to **seventeen**, and — just as importantly — removes the assumptions that made adding a fifth language awkward.

While wiring the new locales up I kept hitting the same shape of bug: code that treated the world as *Chinese or English*. Those spots were invisible with four locales but produce visibly wrong UI as soon as a fifth exists, so they are fixed here rather than left for the next language to trip over.

New locales: `de`, `es`, `fil`, `fr`, `id`, `it`, `ja`, `ko`, `la`, `pt-BR`, `sa`, `ta`, `uk`.

### 1. Catalogs

Each new catalog follows the pattern `ru` already established — `{ ...en, ...overrides }` — so every key always resolves and a newly added upstream key can never render blank while a translation catches up.

Roughly **350 keys per locale** are translated, chosen to cover what a user sees continuously rather than spread thinly: window chrome, sidebar, project menu, connection state, notifications, the permission prompt, the error deck, the agent board, tray copy and the phone layout. Everything else falls back to English, exactly as Russian does today.

I did not attempt all ~6,900 keys per language — machine-bulk translation of the deep settings and Doctor surfaces would be worse than an honest English fallback. The coverage is designed to grow incrementally, and the tests below make growing it safe.

### 2. One source of truth for the locale list

`LOCALES` in `src/i18n/messages/index.ts` now drives the `Locale` type, `isLocale`, and every parity test. Adding a language is one array entry plus a catalog directory.

Locale parsing became table-driven and now handles the tags an OS actually emits — `de-AT`, `es-419`, `fr-CA`, `ta-LK`, `ja_JP.UTF-8`. All Portuguese folds into `pt-BR`; Tagalog (`tl`) shares the Filipino catalog; Indonesian's retired `in` code resolves to `id`.

### 3. Formatting stopped forking on `locale === "zh"`

This was the bulk of the real bug-fixing:

- **25 call sites** sent every non-Chinese locale to `en-US` for dates. A German user would have read English dates in an otherwise German UI. They now go through a new `intlLocale()` helper.
- **Compact counts** were hard-coded to Chinese myriad units. Japanese and Korean also group by 10⁴ but write the units differently (`万/億`, `만/억`), so there is now a per-locale table and a single shared implementation replacing two copies that had drifted apart between `accountUi` and `contextUsage`.
- **Heatmap weekday and month labels** came from two hard-coded arrays; they now come from `Intl.DateTimeFormat`.
- **Durations** were `9小时36分` or `9h 36m` and nothing else; they now use CLDR units.

`intlLocale()` exists because a catalog id is not always safe ICU input: `zh` must become `zh-CN`, and `la` / `sa` have no usable CLDR data so they borrow `en-GB` / `hi-IN`. `htmlLangForLocale` deliberately stays honest (`la` → `la`) — a screen reader must be told the real language even where we borrow formatting data.

### 4. Bugs found and fixed along the way

| Where | Problem |
|---|---|
| `pathRefs.fileSubtitle` | Both call sites collapsed the locale to `locale === "en" ? "en" : "zh"`, so a French or Japanese user saw **Chinese** file-card subtitles. `formatTurnErrorBody` had the same collapsed argument. |
| `session.ts` | `streamFlapCopy` and the unknown-error fallback embedded their English and Chinese text inline instead of using the catalog. |
| `remoteImChannelLabel` | Treated "not Chinese" as "international", so a missing locale defaulted to Chinese brand names; Traditional spellings were missing entirely. |
| `ru` catalog | The `tray.*` keys were translated on the Rust side but **missing from the frontend catalog**, so Russian users saw an English tray menu in-app. Caught by a new test. |

### 5. Native surfaces

`tray_i18n::Locale` gains the thirteen variants, and `parse` / `from_lang_tag` now share one matcher so a settings value and an OS tag cannot disagree about what a locale means. `windows_langid_to_tag` previously mapped only Chinese and English; it now covers the shipped languages, including both Portuguese sublanguages.

`app_menu` had its own four-arm match for the Close label — it now reads `TrayStrings::close`, so a new locale is translated in one place. `session_title` keeps the four hand-tuned native prompts and adds a generic prompt naming the target language for the rest.

### 6. Tests

Catalog tests iterate `LOCALES` instead of naming three locales, plus two new checks that each caught a real regression while writing this:

- always-visible chrome must not still equal the English string in any locale (this is what surfaced the missing Russian `tray.*` keys);
- interpolation placeholders must survive translation — a dropped `{n}` prints a sentence with a hole in it.

`intlLocale` is asserted to produce a tag that `Intl.DateTimeFormat` and `Intl.NumberFormat` actually accept for every locale, so a catalog with thin ICU data can never reach a formatter unguarded.

The existing expectations that asserted `fr` / `de` fall back to English were updated rather than deleted — they now assert the same thing for genuinely unsupported languages (`he`, `th`).

`docs/llm-wiki/i18n.md` documents the formatting helpers and gains a step-by-step "adding a locale" section.

## Verification

| Check | Result |
|---|---|
| `pnpm typecheck` | pass |
| `pnpm lint` | pass, 0 warnings |
| `pnpm test` | 5,990 tests, all i18n suites green |
| `pnpm build:ui` | pass |
| `cargo check` | pass, 0 errors |

Two caveats I would rather state than paper over:

1. **`src/lib/window-config.test.ts` fails on my machine**, and does so on `main` too. It asserts `tray.rs` source text containing `\n`, and my Windows checkout has CRLF line endings. It is unrelated to this PR and should be green on CI.
2. **`cargo test` will not launch on my machine** — the test binary dies at load with `STATUS_ENTRYPOINT_NOT_FOUND`, a local DLL-resolution problem that predates this branch. To avoid shipping unverified Rust, I extracted the pure locale logic into a standalone crate and ran the suite there: **15/15 pass**, covering the round-trip from every catalog id, the Windows LANGID mappings, and that no tray string is empty or has lost its `{pct}` / `{time}` placeholder. Please do give it a normal `cargo test` run in CI.

## Size

Main bundle grows **+266 kB raw / +60.9 kB gzip** (4,850 kB → 5,117 kB raw). That is the cost of thirteen catalogs inlined the same way the existing four are. Happy to move catalogs behind a dynamic `import()` keyed on the resolved locale if you would prefer to pay only for the active language — I left it alone because it changes the loading model for all locales, which felt like your call rather than mine.

## Notes for review

- No version bump and no `CHANGELOG.md` entry — those belong to your release process.
- The Chinese model-facing prompt in `official_aux.rs` is untouched. It is model instruction rather than UI copy, and translating it would change tool-selection behaviour.
- The Chinese keyword heuristics in `ComposerPlusPanel.tsx` and `assistantAnswerSplit.ts` are untouched for the same reason: they match user *input*, not display. Other locales simply do not match them today; extending them is a separate behavioural change.
- I have no way to verify the `la` and `sa` catalogs against a native reviewer, and `fil` / `ta` deserve a second pair of eyes. Please treat those four as the ones most worth checking.


RongleCat (changes requested):
## Review

Thank you — the architectural cleanup is the valuable part: `LOCALES` as the single registry, table-driven OS tag parsing, `intlLocale()` instead of `locale === \"zh\" ? … : en-US`, compact-count tables for CJK, heatmap via `Intl`, and the Russian `tray.*` hole.

We cannot squash-merge the whole 17-locale drop as-is.

### Blocking

1. **CHANGELOG.** Unreleased needs EN + 中文 bullets (Added / Changed as appropriate). `docs/llm-wiki/i18n.md` is updated; the changelog is not.

2. **Drop novelty locales `la` and `sa` from this PR.** Product i18n currently ships `en` / `zh` / `zh-TW` / `ru`. Latin and Sanskrit are not OS UI languages we should advertise in Settings, and they force `intlLocale` fallbacks (`en-GB` / `hi-IN`). Keep the helper so a later locale can borrow CLDR, but do not ship those two catalogs.

3. **Bundle size.** +61 kB gzip for thirteen inlined catalogs is real. Prefer either (a) a smaller first wave of real product languages (`de` `es` `fr` `ja` `ko` `pt-BR` `uk`, plus `it`/`id` if you want) **or** (b) keep all serious locales but load non-active catalogs with a dynamic `import()` so cold start only pays for the resolved language. Do not land Latin/Sanskrit either way.

4. **i18n wiki + Settings picker.** `docs/llm-wiki/i18n.md` must stay the source of truth for shipped ids. The language dropdown must list only locales we are willing to support.

### Please keep

- `intlLocale` / compact-count / heatmap / duration formatting fixes (these already help `ru`).
- Parity tests that iterate `LOCALES`.
- Native tray/menu matcher sharing.

Happy to re-review a slimmer PR: architecture + a first wave of real languages + CHANGELOG, without `la`/`sa`.

PR #717, opened 2026-08-19
feat(pet): stage-reply bubbles, boot color, and peek-hide

oykb58246:
## Summary
Pet chips were title-only and flashed default green on boot. This makes them follow real mid-turn replies, stack per session, and match the requested click / hide / look settings.

- Bubbles appear only when an agent posts a mid-turn assistant reply (not thinking, not the session title). Concurrent chats stack independently and hold until the dismiss timer.
- Overlay boots from saved prefs injected in the init script, so the mark does not flash default green.
- Settings: auto-hide seconds (default 15), optional progress bar (off; spinner / check is enough), six bubble shapes and six backgrounds.
- Single-click opens the main window; double-click hides it; a second click within 3s of opening also hides (quick peek).

## Tests
- `npx vitest run` on pet / i18n / catalog / settings guard files: 137 passed
- `cargo test` compiled `pet_window` (Windows test harness hit `STATUS_ENTRYPOINT_NOT_FOUND` in this environment after a successful compile)

## Notes
Leftover merged locals (`fix/working-list-no-crush`, `fix/working-panel-overlap-ansi`, `fix/premature-turn-done-stuck`, `pr-669-ci`) and `origin/fix/premature-turn-done-stuck` were deleted first. Branch is from official `main`.


RongleCat:
Merged after rebase onto current main (CHANGELOG + pet_window tests kept both Linux overlay and bubble-prefs coverage). Thank you @oykb58246.

PR #720, opened 2026-08-19
fix(pet): wrap bubble body 2-3 lines and keep title readable on light chips

oykb58246:
## Summary
Follow-up to #717. Stage-reply chips were a single ellipsis line, and the session title under them vanished on paper/white (or light-theme) backgrounds because the subtitle color fell back to near-white.

- Body text wraps **2 lines** when a session title is shown, **3 lines** when it is not; the chip grows with the text.
- Session title stays on its own row with contrast-safe muted color on every bubble fill (including paper/white).
- With the progress bar off, the chip no longer keeps an empty track gap.
- Overlay viewport / Rust reserved height updated so taller chips do not clip.

## Tests
- `npx vitest run src/lib/pet/petTasks.test.ts src/lib/pet/petBubbleLayout.test.ts` — 21 passed

## Screenshots
Light chips previously looked like they had a blank second row; that row was the session title. Dark chips already showed it (e.g. 「询问含义是什么」).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ See `docs/llm-wiki/release.md`.
 - **把另一段对话当作上下文加入**：`/attach-chat`、输入框 `+`、侧栏对话图标或右键「加入目前对话」可加入本地对话；chip 挂在输入框；日记只存 `[[chat:id]]`；Host 只给 Agent 加一段压缩记录（最多 3 段；来源对话不改）。拖行身到项目仍是移动会话。
 
 ### Fixed
+- **Pet bubble body wraps 2–3 lines; session title stays readable on light chips**: Stage text is no longer a single ellipsis line. The session title under it uses contrast-safe muted color (paper/white no longer hides it). With the progress bar off, the chip does not keep an empty track gap.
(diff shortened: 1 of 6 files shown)


RongleCat (approved):
## Review

Small, scoped follow-up to #717. Body wraps 2–3 lines, the session title only renders when it differs from the snippet, muted color is contrast-safe on paper/white, and the empty progress-track gap is gone. JS/Rust reserved height stay in lockstep (`PET_BUBBLE_ROW_H=68` → viewport 266).

CI on the PR head was red because the branch was 3 commits behind `main` (unrelated `cargo fmt` plus `session.test.ts` expecting `open_account` after #705). Cherry-picked onto current `origin/main` and verified locally:

- `npx vitest run src/lib/pet/petTasks.test.ts src/lib/pet/petBubbleLayout.test.ts` — 21 passed
- `cargo test --lib pet_window` — 19 passed

Squash-merging.

RongleCat:
已 squash 合入 `main`，谢谢 @oykb58246。浅色气泡上的会话标题现在能看清，正文也可以折 2–3 行。

PR #724, opened 2026-08-19
i18n: first wave of nine product languages (supersedes #708)

1llum1n4t1s:
Supersedes #708, rebuilt against your review. Thank you for the specific blockers — they made the rewrite mechanical.

## Your four blockers

| Blocker | This PR |
|---|---|
| **CHANGELOG** | `## [Unreleased]` gains EN + 中文 bullets under both **Added** and **Fixed**. |
| **Drop `la` / `sa`** | Dropped. I also dropped `fil` and `ta`: they were not on your first-wave list, and they were two of the four catalogs I said I could not have checked by a native reviewer. `intlLocale` keeps the borrow mechanism so a later locale with thin CLDR data can still use it — no shipped locale needs it today. |
| **Bundle size** | Option **(a)**: nine real product languages — `de` `es` `fr` `id` `it` `ja` `ko` `pt-BR` `uk`. Measured against a fresh `main` build: **+170 kB raw / +44 kB gzip** (4,850 → 5,020 kB raw), down from +266 kB / +61 kB. I did not move catalogs behind a dynamic `import()` — that changes the loading model for every locale, and you asked for a smaller first wave rather than a new loading strategy. Happy to do it as a follow-up if you want the cold start back. |
| **Wiki + Settings picker** | `docs/llm-wiki/i18n.md` now states outright that it is the source of truth for the shipped ids, and "Adding a locale" starts by confirming the language is one we intend to support. The picker is no longer a hand-written list: it renders from `LOCALES` against a `Record<Locale, string>` of endonyms, so it **cannot** offer an id we do not ship, and a catalog without a name is a type error rather than a blank row. A test asserts the two agree. |

Everything you asked to keep is kept: `intlLocale`, the compact-count table, the heatmap labels, the duration formatting, the parity tests that iterate `LOCALES`, and the shared native tray/menu matcher.

## What auditing the result turned up

After rebuilding the PR I audited the branch again, which found real bugs — several of them older than this change. They are in the last four commits.

**Crash**

- `clean_llm_title` stripped a surrounding quote pair with `t[1..t.len() - 1]`. `「」` and `“”` are three bytes each, so that slice cut a codepoint in half and panicked — and those are exactly the quotes a title model puts around Japanese and Chinese output. `refine_title_in_background` swallowed the panic, so auto-titling just silently stopped and the session kept its heuristic name. This predates the PR and affects `zh` / `zh-TW` on `main` today. Quotes are peeled by character now, and the 120-unit sanity cap counts characters as well, so a Japanese title is no longer rejected at a length an English one passes.

**Locale never actually reached the user**

- The boot shell discarded every locale except `zh` / `zh-TW` / `en`, so a Japanese desktop painted “Welcome to Grok” and `<html lang="en">` even though the Host had injected the right id. It carries a `BOOT_I18N` table for all thirteen now — the React catalog is not loaded that early, so the splash needs its own copy.
- Follow-system re-resolved from `navigator.language` alone on every mount. A Dock- or WebView2-launched app reports `en-US` on a Japanese desktop, so the locale resolved from the OS was overwritten with English immediately after. It now prefers the Host-injected OS tag, the same one the tray resolves.

**Formatting that still ignored the choice**

- `formatQuotaResetTime` hand-built `MM-DD HH:mm`, and the tray had the same order hard-coded — US month-day order inside an otherwise translated template. The tray takes a per-locale format now.
- Plan history, trace history and the mirror audit log each carried a byte-identical `formatAt` passing `undefined` to `toLocaleString`, which is the WebView language rather than the chosen one. One shared `formatListTimestamp` replaces all three. `locale` is a **required** parameter on these helpers — it was optional, and that is exactly how the account-switcher rows kept silently formatting in English.

**Copy**

- `session.new` / `placeholderTitle` / `untitled` existed only for `en` / `zh` / `zh-TW` / `ru`, so the new locales wrote the English "New chat" **to disk** on first send while the tray said "Neuer Chat". They are translated to match `tray_i18n`, and the always-visible chrome test guards them. Placeholder detection now reads `TrayStrings` for every id in the new public `tray_i18n::ALL` instead of duplicating the names — a hand-kept list stops matching the moment a catalog is translated, which strands those sessions on their default name forever. Remote IM's own two-entry copy delegates to it; it never recognized `zh-TW`.
- German rendered "turn" as **"Zug"**, which is a train or a chess move. Notifications, the error deck and the Remote IM stop reply read as if the app were reporting on rail traffic. All seven strings use "Vorgang". The other catalogs check out: `fr` *tour*, `it`/`es` *turno*, `id` *giliran*, `zh` 轮.
- `AGENTS.md` still promised seventeen locales including `fil` `la` `sa` `ta`, contradicting both the code and the wiki.

## Verification

| Check | Result |
|---|---|
| `pnpm typecheck` | pass |
| `pnpm lint` | pass, 0 warnings |
| `pnpm test` | **6,004 / 6,004 pass** |
| `pnpm build:ui` | pass |
| `cargo check` | pass, 0 errors |
| `cargo check --tests` | pass, 0 errors |
| `cargo test` (see below) | **1,300 / 1,300 pass** |

**On running the Rust tests locally.** `cargo test` aborts at process start on Windows with `STATUS_ENTRYPOINT_NOT_FOUND`, which is why the note in `build.rs` sends this to CI. The specific cause is `comctl32.dll :: TaskDialogIndirect` (pulled in by `rfd`): only comctl32 **v6** exports it, System32 ships v5.82, and the test harness gets no manifest to activate the v6 assembly. Reproducing your CI steps by hand — `cargo test --no-run`, `mt.exe` embedding `windows-test-manifest.xml`, then running the harness directly — makes the suite run: **1,300 passed, 0 failed**.

That was worth doing, because it immediately caught a bug in this branch that CI cannot see: `engine_lang_follows_product_locale_not_hardcoded_zh` asserted `matches!(lang, "en" | "zh" | "zh-TW")`, so once `app_locale()` resolves the new languages it panics with `unexpected catalog id ja` on a Japanese desktop while staying green on an English runner. It holds `APP_HOME_ENV_LOCK`, so that one panic poisoned the mutex and produced 26 failures. It now checks against `tray_i18n::ALL` (last commit).

I still ran the pure locale and title logic in a standalone crate as a second check — **18/18**, including a `#[should_panic]` test that reproduces the old byte slice to confirm the CJK crash was real.

`src/lib/window-config.test.ts` also fails on my Windows checkout on `main` (it asserts `tray.rs` source text containing `\n` and my working copy has CRLF); it passes here only because this branch rewrites that file. Unrelated either way.

## Notes for review

- **A flake you may want to know about, unrelated to this PR:** the suite is not parallel-safe on a machine that already has app data. `plan_chrome::tests::upsert_and_mark_stale_round_trip` repoints `GROK_APP_HOME` under a module-private `ENV_LOCK` instead of `paths::APP_HOME_ENV_LOCK`, so it moves the app home out from under whichever suite is holding the real lock. `integration_test::project_add_trust_remove_roundtrip` then reads a project list written to a different home and fails its assertion *while holding* that lock — and because most acquisitions are `.lock().unwrap()` rather than the `unwrap_or_else(|e| e.into_inner())` that `with_isolated_project_store` already uses, one race becomes ~28 `PoisonError` failures. Measured here: 3/3 serial runs green, 2/3 parallel runs red; with `plan_chrome` moved onto the shared lock and the acquisitions made poison-tolerant, 3/3 parallel runs green. I left it out since it is neither mine nor in scope — happy to send it as its own PR.
- No version bump and no release-notes edit — those belong to your process.
- The Chinese model-facing prompt in `official_aux.rs` is untouched: it is model instruction, not UI copy, and translating it would change tool-selection behaviour. Same reasoning for the Chinese keyword heuristics in `ComposerPlusPanel.tsx` and `assistantAnswerSplit.ts` — they match user *input*, not display.
- **Known gap I did not close:** Telegram's `setMyCommands` still registers only an English default plus a Chinese list, so a `ja` / `de` / `ko` client sees an English command palette. Closing it means 16 commands × 11 more locales and one `setMyCommands` round-trip per language on bridge start — Telegram only accepts two-letter ISO 639-1 codes, so `pt-BR` folds to `pt` and `zh-TW` cannot be targeted separately. `slash.rs` and `channels/telegram.rs` are untouched by this PR, and it felt like the wrong thing to bolt onto a change you asked to slim down. I have since written it, so it is ready whenever you want it.
- The native menu bar (`File` / `Edit` / `View` / `Window` / `Help`) is still English. muda 0.19.3 hard-codes English for every `PredefinedMenuItem` on all platforms (`items/predefined.rs`), so translating only the submenu titles would put German headings over English items. Doing it properly means passing a label to every predefined item — 19 strings per locale; also a separate change, and also already written if you want it.


1llum1n4t1s:
Closing this — you shipped the fifteen catalogs yourself in `feat(i18n): ship 15 complete UI locales`, which is the whole first half of this PR, and every catalog file here now conflicts add/add with yours. Yours wins; mine has nothing to add there.

What is still missing is the correctness half, so I re-cut it against your `main` and split it by concern:

- **#725** — `main` is red on every job: Windows has no build at all (`win_crash.rs` against `windows` 0.61), `cargo fmt --check` blocks all three Rust jobs before they start, and `session.test.ts` still asserts the pre-#705 `AUTH_NO_CONTEXT` deck.
- **#726** — three bugs the rollout turned on: `clean_llm_title` panics on `「」` / `“”` titles, `is_placeholder_title` recognises no locale added since `ru` so those sessions are never auto-titled, and `engine_lang_follows_product_locale_not_hardcoded_zh` fails 37 tests on a non-English desktop.

Three more are written and verified locally, waiting on whether you want them: the boot shell and follow-system resolution (a `ja` desktop still paints English on first frame), the date/time helpers that ignore the chosen locale, and the German catalog, which renders "turn" as "Zug" — a train — in about seventy places. Say the word and I will open them.

Thanks for the review on #708; dropping `la` / `sa` was right, and so was cutting the scope.

PR #727, opened 2026-08-19
fix(test): make cargo test parallel-safe again

1llum1n4t1s:

`cargo test` without `--test-threads=1` is not reliable, and when it fails it hides why.

## The race

`GROK_APP_HOME` is process-global. Every suite that repoints it holds `paths::APP_HOME_ENV_LOCK` while it does — except `plan_chrome::tests`, which declares its own `static ENV_LOCK: Mutex<()>`. That serialises `plan_chrome.rs` against itself and nothing else, so it moves the app home out from under whichever suite is holding the real lock.

Which test dies depends on timing:

- `plan_chrome::tests::upsert_and_mark_stale_round_trip` writes its chrome to one home and reads from another → `expect("chrome saved")`.
- `integration_test::integration::project_add_trust_remove_roundtrip` asserts against a project list written to a different home.

## The amplifier

Whichever one loses, it panics **while holding** `APP_HOME_ENV_LOCK`. 38 of that lock's acquisitions were `.lock().unwrap()`, so the poisoned lock takes the rest of the suite with it — one race becomes a wall of `PoisonError` with the real failure buried near the top.

The lock guards a `()`. It serialises an environment variable, so a panic under it leaves no invariant broken and poisoning carries no information. `with_isolated_project_store` and `store::tests` already did `unwrap_or_else(|e| e.into_inner())`; this makes it uniform across all 38.

## Measured

On `main`:

| | |
|---|---|
| `--test-threads=1` | 1,372 / 1,372 with #726 |
| default threads | 1,371 / 1,372 — one genuine race failure |

After:

| | |
|---|---|
| default threads, 3 consecutive runs | **1,372 / 1,372 each** |

Two commits, one per problem — the lock fix alone is enough to stop the race; the poison-tolerance commit is what makes the *next* one legible.

Measured on `main` @ c23c17ed with #726 applied — without it the engine locale assertion fails on a non-English desktop and poisons this very lock, which buries the race under 37 `PoisonError`s. #726 does not touch any file here.

### DIFF
--- a/src-tauri/src/agent_config_edit.rs
+++ b/src-tauri/src/agent_config_edit.rs
@@ -832,7 +832,7 @@ two_pass_compaction = true
 
     #[test]
     fn path_scope_is_agent_home() {
-        let _lock = crate::paths::APP_HOME_ENV_LOCK.lock().unwrap();
+        let _lock = crate::paths::APP_HOME_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
         let p = agent_config_toml();
         assert!(is_agent_home_config_path(&p));
         let other = PathBuf::from("/tmp/not-agent-home/config.toml");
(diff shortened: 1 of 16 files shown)


RongleCat (approved):
Root cause is right: `plan_chrome::tests` used a private ENV_LOCK while every other suite serialises `GROK_APP_HOME` on `paths::APP_HOME_ENV_LOCK`, so parallel `cargo test` stole the app home. Recovering a poisoned `()` lock is also correct — it only guards an env var.

Squash-merging. Thank you.

PR #728, opened 2026-08-19
i18n(de): a turn is a "Vorgang", not a "Zug"

1llum1n4t1s:

The German catalog translates the agent **turn** as **"Zug"** in 125 places. In German a `Zug` is a train, or a move in chess — it is never a unit of work. A German user currently reads:

| key | today | |
|---|---|---|
| `chat.turnFailed` | "Dieser Zug ist fehlgeschlagen" | *This train has failed* |
| `agent.streamStallCancel` | "Zug beenden" | *End the train* |
| `notify.turnDoneTitle` | "Grok hat einen Zug abgeschlossen" | *Grok completed a chess move* |
| `attachChat.scopeRecent` | "Letzte 16 Züge" | *Last 16 trains* |
| `account.col.turns` | "Züge" | *Trains* (a table column) |

## The replacement

**"Vorgang"** — an operation, a process. It is masculine, exactly like `Zug`, so every surrounding article and case ending stays correct with no further edits:

```
dieser Zug   -> dieser Vorgang
diesem Zug   -> diesem Vorgang
des Zugs     -> des Vorgangs
die Züge     -> die Vorgänge
den Zügen    -> den Vorgängen
```

That gender match is why I did not pick `Runde` (feminine) or `Antwort` (feminine) — either would mean touching the words around all 125 sites, and every one of those is a chance to introduce a new mistake.

Compounds follow: `Agentenzug` → `Agentenvorgang`, `Nutzerzug` → `Nutzervorgang`, `Zugende` → `Vorgangsende`, `Mitten-im-Zug` → `Mitten-im-Vorgang`.

## What is deliberately untouched

Ordinary German words that merely contain the letters, and mean something else entirely: `Zugriff` and its compounds (`Netzwerkzugriff`, `Werkzeugzugriff`, `Projektschreibzugriffe`, `Dateizugriff`, `Mikrofonzugriff`), `bevorzugen`, `hinzugefügt`, `freizugeben`, `zugehörig`, `einzugreifen`, `Auszug` / `Auszüge`. The pass asserts their counts are unchanged, and separately asserts no word meaning "turn" survives.

## The other catalogs are fine

I checked `chat.turnFailed` in every shipped locale before touching German:

| | | | | |
|---|---|---|---|---|
| `fr` *tour* | `it` *turno* | `es` *turno* | `id` *giliran* | `pt-BR` *turno* |
| `ru` *ход* | `uk` *хід* | `zh` 轮 | `ja` ターン | `ko` 턴 |

Only German picked a word that changes the meaning.

## Verification

Regenerated on `main` @ c23c17ed rather than rebased, so the pass ran against the German catalog as it stands after #721 — same 125 sites, same rules. `pnpm typecheck` and `pnpm lint` pass; `pnpm test` is **6,165 / 6,166**, the same as `main` itself (the one failure is the CRLF assertion in `window-config.test.ts`, sent as #730).

**A native German speaker should sanity-check the word choice.** I am confident `Zug` is wrong and that `Vorgang` is grammatically safe; whether `Vorgang` or `Durchgang` reads better in your product voice is a call I would rather you made. The change is one word, so swapping it later is a single find-and-replace.


RongleCat (approved):
Agreed that *Zug* is the wrong word for an agent turn (train / chess move). *Vorgang* keeps gender and case endings, so the 125-site swap is mechanically safe. `Zugriff` / `bevorzugen` / `Auszug` and the rest of the exclusion list are left alone.

A native-speaker polish pass can still swap *Vorgang* for *Durchgang* later; this is a single-token change.

Squash-merging. Thank you.

PR #729, opened 2026-08-20
fix(i18n): dates and clocks ignore the chosen language

1llum1n4t1s:

Six places print a date or a clock. All six ignore the language the user picked in Settings.

| where | today | should be (de) |
|---|---|---|
| Plan history rows | `Apr 15, 2026, 09:05` | `15. Apr. 2026, 09:05` |
| Trace export rows | `Apr 15, 2026, 09:05` | `15. Apr. 2026, 09:05` |
| Mirror write-ACL audit | `Apr 15, 2026, 09:05` | `15. Apr. 2026, 09:05` |
| Plan history preview header | `4/15/2026, 9:05:00 AM` | `15.4.2026, 09:05:00` |
| Quota reset (user menu, saved accounts, account panel) | `04-15 09:05` | `15.04. 09:05` |
| Quota reset (native tray usage row) | `04-15 09:05` | `15.04. 09:05` |

## Why they were wrong

The first four called `toLocaleString(undefined, …)`. `undefined` does not mean "the app's language" — it means whatever the WebView reports, which is the OS language on some launch paths and `en-US` on others. Switching the app to German never reaches it.

The last two never asked a formatter at all: both hand-assembled `MM-DD HH:mm`. That string is a date only to a reader who already expects month first. A German user parses `04-15` as the 4th of month 15, and an en-US user never gets the am/pm their clock format needs.

## What changed

**One helper instead of four copies.** `PlanHistoryList`, `TraceHistoryList` and `MirrorConnectPanel` each carried a byte-identical `formatAt`/`formatExportedAt`/`formatAuditAt`, and `AppWorkbench` inlined a fourth. They are now one `formatListTimestamp(iso, locale)` in `src/lib/formatDateTime.ts`, built on the existing `intlLocale()` so `zh` resolves to `zh-CN` rather than the macrolanguage.

**`locale` is a required prop**, not optional. Optional would have left every call site that forgot it on the old behaviour, silently — which is the bug this PR is fixing. `UserMenu` had no locale at all and gains one; `AccountPanel`, `RuntimeSection` and `RemoteImSection` already had one in scope.

**The tray format lives in `TrayStrings`.** `reset_time_fmt` travels with the rest of that locale's tray copy, so adding a locale means filling one more field in a table the compiler already forces you to complete — not remembering that a `chrono` format string exists in another file. `%p` appears only for `en` and `fil`: chrono spells it `AM`/`PM` in English with no localized form, so a 12-hour clock would be wrong in the other thirteen.

## Verification

`pnpm typecheck` and `pnpm lint` pass. `pnpm test`: **6,166 / 6,166** on a checkout with #730 applied, or 6,165 / 6,166 without it — the one failure there is the CRLF assertion #730 fixes, in a file this PR does not touch.

`cargo test`: **1,373 / 1,373** at default threads, up one from the new locale test. Measured on `main` @ c23c17ed with #726 and #727 applied — without them `main` fails 38 tests on a non-English desktop before reaching anything here. Neither touches these files.

New tests: `formatDateTime.test.ts` (4) asserts the tag really reaches `Intl` rather than a template — `zh` and `zh-TW` must not collapse onto each other. `accountUi.test.ts`'s reset-time case now asserts `de` is day-first and `en` keeps 12-hour, instead of pinning the old hand-built shape. `tray_i18n.rs` renders all 15 formats, because `chrono` panics at `Display` time on a bad spec rather than at compile time.

## Note on the date order table

I picked each locale's short-date shape from the usual convention for that language (day-first for `de` `es` `fr` `id` `it` `pt-BR` `ru` `ta` `uk`, month-first for `en` `fil` `ja` `zh` `zh-TW`, and `ko` as `04. 15.`). The JS side does not need this — `Intl` knows — but the tray has no `Intl`, so the table is hand-written and worth a glance from anyone who reads one of those languages natively.

### DIFF
--- a/src-tauri/src/tray.rs
+++ b/src-tauri/src/tray.rs
@@ -139,28 +139,27 @@ pub fn build_menu(app: &AppHandle) -> Result<Menu<Wry>, tauri::Error> {
     builder.build()
 }
 
-/// Format quota refresh ISO → local `MM-DD HH:mm` (same as in-app user menu).
-fn format_reset_mm_dd_hm(iso: &str) -> Option<String> {
+/// Format quota refresh ISO → local short date and clock, in `fmt`.
+///
+/// `fmt` comes from the active locale's `TrayStrings`, so this stays the same
+/// instant the in-app user menu shows, written the way that language writes it.
+fn format_reset_time(iso: &str, fmt: &str) -> Option<String> {
     let s = iso.trim();
     if s.is_empty() {
         return None;
     }
     // Prefer RFC3339 (what BillingSnapshot writes).
     if let Ok(dt) = chrono::DateTime::parse_from_rfc3339(s) {
-        return Some(
-            dt.with_timezone(&chrono::Local)
-                .format("%m-%d %H:%M")
-                .to_string(),
-        );
+        return Some(dt.with_timezone(&chrono::Local).format(fmt).to_string());
     }
     // Fallback: UTC Z without offset, or naive local.
     if let Ok(ndt) = chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%S%.fZ") {
         let dt = ndt.and_utc().with_timezone(&chrono::Local);
-        return Some(dt.format("%m-%d %H:%M").to_string());
+        return Some(dt.format(fmt).to_string());
     }
     if let Ok(ndt) = chrono::NaiveDateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%SZ") {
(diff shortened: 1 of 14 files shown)


RongleCat (approved):
Correct diagnosis: `toLocaleString(undefined, …)` is the WebView language, not Settings, and the hand-built `MM-DD HH:mm` is month-first for every locale. One `formatListTimestamp` + required `locale` prop is the right shape (optional would have silently kept the bug). Tray `reset_time_fmt` on `TrayStrings` is the right place so a new locale cannot forget it.

Squash-merging. Thank you.

PR #731, opened 2026-08-20
fix(i18n): the app boots in the wrong language, and "follow system" does not

1llum1n4t1s:

Set the language to Japanese, restart, and the first thing on screen is `Welcome to Grok`. Set it to "Follow system" on a Japanese Windows box and the app stays English forever.

Three separate causes, one per commit.

## 1. "Follow system" read the WebView, not the system

`applySystem` resolved the locale from `navigator.language`. That is the WebView's language, not the OS's — WebView2 reports `en-US` on a Windows install whose UI language is Japanese. So the setting did nothing, and changing the system language did nothing either.

`readSystemLangTag()` already exists for exactly this and already prefers the OS tag the Host injects as `__GROK_BOOT_OS_LANG__`. This one call site simply never used it.

The `languagechange` handler keeps reading the navigator: that event means the WebView's own list moved, which a tag captured at window build time cannot know about.

## 2. `<html lang>` was `en` for twelve of fifteen locales

The Host sets `<html lang>` before first paint with:

```rust
let boot_html_lang = match boot_locale_tag {
    "zh" => "zh-CN",
    "zh-TW" => "zh-TW",
    _ => "en",
};
```

Every catalog except the two Chinese ones fell into that `_`. `<html lang>` is not decoration — it selects the font fallback (a shared CJK codepoint renders with the wrong regional glyph when the tag is wrong), the hyphenation dictionary, and how a screen reader pronounces the page.

Now `Locale::html_lang()`, mirroring `htmlLangForLocale` in `src/i18n/index.ts`, so the mapping sits next to the ids it maps and a new locale carries it automatically.

## 3. The splash spoke three languages out of fifteen

`index.html` paints before the JS bundle arrives, so its copy is a literal table. That table held `zh`, `zh-TW` and `en`. Anything else was **discarded — including a perfectly good `ja` the Host had already injected** — and re-derived from the navigator, which collapsed everything non-Chinese to English:

```js
var loc = window.__GROK_BOOT_LOCALE__;
if (loc !== "zh" && loc !== "zh-TW" && loc !== "en") {
  /* ignore what the Host said, guess from navigator, end up at "en" */
}
```

All fifteen are now filled in, taken verbatim from each catalog's `setup.title` and `setup.detecting` — the same two strings React shows a moment later when it swaps `#root` for the setup gate. `src/i18n/bootShell.test.ts` asserts they stay equal, because a table that drifts shows the user two different sentences within the same second.

The navigator fallback (browser `vite dev`, or a settings id the build predates) now walks the same table rather than collapsing to English, with the `pt` / `tl` / `in` aliases `normalizeLocale` already handles.

## Verification

`pnpm typecheck` and `pnpm lint` pass. `cargo test`: **1,373 / 1,373** at default threads, up one from the new `html_lang` test. Measured on `main` @ c23c17ed with #726 and #727 applied — without them `main` fails 38 tests on a non-English desktop before reaching anything here; neither touches these files.

`pnpm test`: **6,166 / 6,166** on a checkout with #730 applied. Without it the count is 6,165 / 6,166 — that one failure is `window-config.test.ts` asserting `tray.rs` source against `\n`, which is red on any Windows checkout with `core.autocrlf=true` and is what #730 fixes. Nothing to do with i18n, so it is its own PR rather than folded in here.

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,42 +245,63 @@
         } catch (e) {
           /* ignore */
         }
+        /*
+         * Boot copy for every shipped catalog. The splash paints before the JS
+         * bundle arrives, so it cannot read src/i18n — src/i18n/bootShell.test.ts
+         * asserts each row still equals that locale's `setup.title` and
+         * `setup.detecting`, so the two cannot drift apart.
+         */
+        var BOOT_COPY = {
+          "de": ["Willkommen bei Grok", "Grok Build wird geprüft…"],
+          "en": ["Welcome to Grok", "Checking Grok Build…"],
+          "es": ["Te damos la bienvenida a Grok", "Comprobando Grok Build…"],
+          "fil": ["Maligayang pagdating sa Grok", "Sini-check ang Grok Build…"],
+          "fr": ["Bienvenue dans Grok", "Vérification de Grok Build…"],
+          "id": ["Selamat datang di Grok", "Memeriksa Grok Build…"],
+          "it": ["Benvenuto in Grok", "Verifica di Grok Build…"],
+          "ja": ["Grok へようこそ", "Grok Build を確認中…"],
+          "ko": ["Grok에 오신 것을 환영합니다", "Grok Build 확인 중…"],
+          "pt-BR": ["Boas-vindas ao Grok", "Verificando o Grok Build…"],
+          "ru": ["Добро пожаловать в Grok", "Проверка Grok Build…"],
+          "ta": ["Grok க்கு வரவேற்கிறோம்", "Grok Build ஐச் சரிபார்க்கிறது…"],
+          "uk": ["Ласкаво просимо до Grok", "Перевірка Grok Build…"],
+          "zh": ["欢迎使用 Grok", "正在检测 Grok Build…"],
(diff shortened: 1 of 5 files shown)


RongleCat (approved):
Three independent boot-locale bugs, all real:

1. \"Follow system\" reading `navigator.language` ignores the OS tag Host already injects as `__GROK_BOOT_OS_LANG__`. Using `readSystemLangTag()` is the existing API this call site missed.
2. Host `<html lang>` mapping `_ => \"en\"` marked twelve catalogs as English. `Locale::html_lang()` matching `htmlLangForLocale` is the right SoT.
3. Splash copy table only knew en/zh/zh-TW and discarded a good Host `ja`/`de`. Filling all 15 from `setup.title` / `setup.detecting` plus `bootShell.test.ts` keeps the splash and the setup gate in lockstep.

Squash-merging. Thank you.

RongleCat:
Integrated onto `main` as 319a4d86 (squash of the three commits). Conflict with #726/#729 in `tray_i18n.rs` was both tests kept: the reset-clock renderer from #729 and `html_lang` from this PR.

Thank you — this is the difference between a Japanese Windows box booting in English and booting in Japanese.

PR #736, opened 2026-08-20
i18n(ja): translate the Japanese catalog (4,690 strings are still Simplified Chinese)

### LINKED ISSUE #732: [Bug]: the Japanese UI is 67% Simplified Chinese
## What happens

Switch the language to 日本語 and about two thirds of the UI is Simplified Chinese. Settings is almost entirely Chinese — a Japanese user sees a "日本語" option that produces a Chinese app.

`src/i18n/messages/ja/settings-remoteIm.ts` on `main` @ c23c17ed, top of file:

```ts
/** ja messages — domain: settings-remoteIm */
export const jaSettingsRemoteIm = {
  "settings.remoteIm.sidebarAria": "IM 通信渠道",
  "settings.remoteIm.bridgeOverview": "Bridge 总览",
  "settings.remoteIm.group.domestic": "国内",
  "settings.remoteIm.conn.websocket": "WebSocket 长连接 · 无需公网",
  "settings.remoteIm.comingSoon": "即将支持",
  "settings.remoteIm.retiredTitle": "该渠道已不再支持",
```

Every one of those is the `zh` string verbatim.

## How much

**4,690 of 7,000 keys — 67%.**

| file | keys | still Chinese | |
|---|---:|---:|---:|
| `ja/settings-remoteIm.ts` | 635 | 568 | 89% |
| `ja/extensions.ts` | 641 | 527 | 82% |
| `ja/settings-ui.ts` | 540 | 526 | 97% |
| `ja/settings.ts` | 489 | 444 | 91% |
| `ja/doctor.ts` | 491 | 387 | 79% |
| `ja/settings-agent.ts` | 380 | 339 | 89% |
| `ja/session.ts` | 327 | 312 | 95% |
| `ja/composer.ts` | 363 | 285 | 79% |
| `ja/automations.ts` | 281 | 279 | 99% |
| `ja/settings-code.ts` | 199 | 197 | 99% |
| `ja/providers.ts` | 182 | 171 | 94% |
| `ja/workspace.ts` | 515 | 160 | 31% |
| `ja/settings-memory.ts` | 154 | 149 | 97% |
| `ja/account.ts` | 158 | 135 | 85% |
| `ja/slash.ts` | 113 | 111 | 98% |
| `ja/settings-pet.ts` | 79 | 62 | 78% |
| `ja/features.ts` | 520 | 15 | 3% |
| `ja/core.ts` | 299 | 13 | 4% |
| `ja/tasks.ts` | 244 | 6 | 2% |
| `ja/chat.ts` | 158 | 3 | 2% |
| `ja/sidebar.ts` | 89 | 1 | 1% |
| `ja/errors.ts`, `ja/kanban.ts`, `ja/project.ts` | 143 | 0 | 0% |
| **total** | **7000** | **4690** | **67%** |

The shape suggests the chat surfaces were translated and the settings surfaces were not.

## How this was counted

Japanese and Chinese share characters, so "contains Han" would over-count. The test used instead is:

> the value is byte-identical to `zh` **and differs from `zh-TW`**

A word Japanese genuinely shares with Chinese (`保存`, `最小化`) is written the same in *both* Chinese catalogs, so it never matches. A value that matches `zh` but not `zh-TW` contains simplified-specific characters — `无`, `设`, `渠`, `话` — which modern Japanese does not use at all. That is 4,174 keys. A further 516 match both Chinese catalogs and are Chinese rather than Japanese in context (`通道`, `已用`, `重置`, `峰值 Token`, `模型`, `{count} 天` — Japanese would be チャネル / 使用済み / リセット / ピーク / モデル / {count}日), giving 4,690.

Excluded as deliberately Chinese in every catalog: `settings.sttLanguage.zhCN`, `settings.sttLanguage.zhTW`, `settings.terminalFontPreview`, and the `prov.preset.volcanoArk.*` product name.

## The other twelve catalogs are clean

Same scan across `de es fil fr id it ko pt-BR ru ta uk`: **zero** leaked values. `ko` in particular is fully Korean. This is only `ja`.

One smaller thing in the same family, which I can send as a PR whenever it is useful: `ja/session.ts` has `"session.placeholderTitle": "新会话"`, while `tray_i18n.rs`'s `JA` correctly says `"新しいチャット"`. So the native tray and the stored session title disagree for Japanese even after the catalog is fixed.

## Worth guarding

Whatever fills these in, the check above is cheap enough to run in CI and would have caught this before release — a catalog value that equals `zh` and differs from `zh-TW` is never correct outside the excluded keys. Happy to send that as a test alongside a fix if you want it.

I can take the translation itself if that is useful to you — say the word and I will open it. I am not opening a 4,700-string PR unasked, because you have `scripts/i18n-merge-fill.py` and may well prefer to run your own pipeline over it.

1llum1n4t1s:
Fixes #732.

The `ja` catalog shipped with the 15-locale rollout is mostly untranslated: 4,690 of its 7,004 values are still the Simplified Chinese source text. A Japanese user gets a Chinese UI everywhere except chat and errors.

This translates all 4,690 of them.

## How the 4,690 were identified

A value is counted as leaked source only when it is **identical to `zh` and different from `zh-TW`**. That discriminator matters: "contains Han characters" over-counts badly, because Japanese shares kanji with Chinese, and because some values are legitimately Chinese in every locale (language names like `简体中文`, the font preview, product names like `火山方舟`). Those five keys are excluded explicitly.

The same scan over the other twelve catalogs finds zero leaked values, so this is a `ja`-only problem.

## What changed

20 files under `src/i18n/messages/ja/`, 4,566 lines. The remaining 124 of the 4,690 are words Japanese and Chinese write identically (`保存`, `内容`, `国内`), so their lines are unchanged.

Nothing outside `src/i18n/messages/ja/` is touched — no key added, renamed or removed, no structural edit. Every value was replaced in place by a script that refuses to run if a `{placeholder}` or a `\n` / `\"` escape would be lost, so the catalog shape is provably identical to before.

## Translation conventions

Taken from the ~2,310 keys in `ja/chat.ts` and `ja/errors.ts` that were already Japanese, so the new text reads like the old:

- 体言止め for labels and buttons, 敬体 for hint sentences
- half-width space around Latin runs — `CLI のパスを設定`, `構造化 JSON`
- `…` for progress, `—` for the hint dash, `「」` for quoted UI names
- product and protocol names kept in Latin: `Bridge`, `Socket Mode`, `agent-home`, `soft-fail`, `--permission-mode`
- Chinese-only channel names rendered as the brands Japanese users know: `飞书` → `Feishu`, `钉钉` → `DingTalk`, `企业微信` → `WeCom`, `微信个人` → `WeChat（個人）`, `微博` → `Weibo`

The honesty copy that runs through this catalog — "unset is not off", "never invents", "soft-fail is not success" — is translated as honesty copy, not softened.

## Verification

- `pnpm typecheck` — clean, and re-run after each of the eight commits
- `pnpm lint` — clean
- `pnpm test` — 6,165 / 6,166. The one failure is `src/lib/window-config.test.ts`, which asserts on the raw bytes of `src-tauri/src/tray.rs` and so fails on any Windows clone with git's default `core.autocrlf=true`. It fails identically on a pristine `main` and is unrelated to this change (#730 fixes it).
- the leak scan re-run after the last commit reports 0 remaining

## Review note

This is a large single-locale diff and it is fine to land it in pieces — the eight commits are self-contained and each one typechecks on its own.


1llum1n4t1s:
Closing — #733 landed this first, and it landed it better.

I re-ran my scan against `main` at 253cccb6 to check for gaps rather than assume: 25 values still match `zh` and differ from `zh-TW`, and they are exactly the shared-shinjitai set your PR body already calls out — `保存`, `独立`, `内容`, `国内`, `宝石`, `斜体`. Those are correct Japanese, not leaked source. There is nothing left for this PR to add, so re-translating 4,690 already-correct strings would be pure churn.

The catalog test you added in #733 is the part I did not think to build, and it is the more valuable half: my diff fixes `ja` once, that test makes the next locale rollout fail loudly instead of shipping quiet Chinese.

Thanks for turning #732 around this fast.

PR #751, opened 2026-08-20
fix(i18n): two translated strings dropped their {name} placeholder

1llum1n4t1s:
`t()` substitutes by name, so a translation that drops `{name}` renders without the value. It is not a missing key and not an empty string, so none of the existing catalog checks sees it. Two strings shipped that way.

### 1. Japanese PR hub shows no author

`GitPrHubPanel.tsx` calls `tr("prHub.author", { name: pr.author })`, but `ja` reads `"作成者"` with no placeholder, so every row says "created by" and stops. All fourteen other locales keep `{name}`:

| | |
|---|---|
| en | `by {name}` |
| ja | `作成者` ← |
| ko | `{name} 작성` |
| zh | `作者 {name}` |
| de | `von {name}` |

### 2. Ukrainian ships a command with no argument

`settings.agentsPersonas.preferredCurrent` renders a command the user is meant to copy. `uk` lost the `{name}` inside the backticks:

```
en: spawn `--agent {name}` (soft-respawn when live).
uk: запуск `--agent ` (м'який перезапуск, коли живе).
```

### Regression guard

`messages.test.ts` already spot-checks six keys for their placeholders. This generalises that into a catalog-wide sweep: every locale must carry exactly the placeholders its `en` string declares — dropped, added, or renamed all fail. The two bugs above are the only strings in the fifteen catalogs that fail it, so the sweep goes green with this fix and stays that way.

Both strings predate the recent i18n work (`06a56956` and `574d7f18`); `#732` re-translated `ja` but did not touch this line.

### Verification

`pnpm typecheck`, `pnpm lint`, `pnpm test` (451 files, 6176 tests) all green. No Rust touched. I also confirmed the guard is not vacuous by reintroducing the `ja` bug and watching it fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ See `docs/llm-wiki/release.md`.
 - **宠物换成完整 bloub 形变目录**：浮层改用测过的 SVG 引擎（静止、思考、眨眼、通知、警示、六边形、轨道、彗星等），不再用旧的 clip-path 表情。设置可选 8 种休息形体和 16 种休息表情。输入框打字会走目录里的警示（斜感叹号），停打后回到所选休息形体；未读完成会话是显眼的橙红圆点（身体已经很热时改用柠黄），不再用视频里的蓝色。
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


RongleCat:
Thanks @1llum1n4t1s — restoring `{name}` is the right fix, and the catalog-wide placeholder sweep will catch this class of bug next time. Added Unreleased CHANGELOG (en+zh). Squash-merging now.

Issue #754, opened 2026-08-20
[Bug]: App freezes at end of turn — the UI thread blocks on a lock inside the window procedure


1llum1n4t1s:
## Describe the bug

The window stops responding when a long turn finishes. The Rust backend stays perfectly healthy — the heartbeat keeps ticking, the loopback servers keep listening, nothing panics — but the window never repaints or takes input again.

I captured a full user-mode dump of the frozen process and walked the stacks. The main/UI thread is blocked **inside a window procedure**:

```
ntdll!ZwWaitForAlertByThreadId      <- current position
ntdll!RtlWaitOnAddress
KERNELBASE!WaitOnAddress            <- SRWLock / futex wait (Rust Mutex / Condvar)
grok-app.exe+0x1eacb2f
grok-app.exe+0x2a7b268
comctl32!DefSubclassProc            <- still inside the WndProc
grok-app.exe+0x1b8d630
user32!CallWindowProcW
user32!SendMessageW
```

A window message arrives, the subclass WndProc takes a Rust lock, and the lock never comes free. The message pump therefore never returns, which is precisely what Windows reports as `Responding = False`.

The wait is `WaitOnAddress`, **not** ALPC / COM (`NtAlpcSendWaitReceivePort`), so this is an in-process lock — not a WebView2 round trip.

### Thread picture (29 threads)

| Thread | State | Reading |
| --- | --- | --- |
| main / UI | `WaitOnAddress` **inside `DefSubclassProc`** | the freeze |
| 5 workers | identical park stack (`+0x18fa5be → +0x1b61fa6 → +0x23ff0a8`) | normal tokio idle |
| one thread **created in the same second as the freeze** | `WaitForSingleObject` | likely the other half of the cycle |
| one more | lock wait at a different site (`+0x1b7ee22`) | possibly involved |
| rest | `GetQueuedCompletionStatusEx`, `NtUserGetMessage`, IME / shell / Google Drive DLLs | unrelated |

Nothing is spinning: the host process burned **0.00 s of CPU over a 3 s sample**, and the WebView renderer sat at ~10% of one core. This is a wait, not a loop.

A background task holding the lock while waiting on the main thread would close an AB-BA cycle and explain everything observed, but I cannot prove which lock it is (see *Symbols* below).

### What the log shows

The last lines before the app went silent forever:

```
INFO session_manager::events: acp prompt_complete deferred stop=end_turn tools=0 perm=false plan=false ask=false
[...]

PR #756, opened 2026-08-20
fix(composer): paste Explorer files (.dmp) via native clipboard paths

### LINKED ISSUE #755: [Bug]: Windows paste of Explorer files (.dmp) fails with NotReadableError
## Describe the bug

On Windows, copying a `.dmp` (crash dump) in Explorer and pasting into the composer fails. The banner shows:

> 无法附加粘贴的文件 (The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired. NotReadableError)

Same failure for other Explorer-copied files whose WebView `File` blob cannot be read (locked dumps, large binaries). Drag-drop of the same path still works because Tauri supplies a native path.

## Steps to reproduce

1. Start Grok App on Windows.
2. In Explorer, copy a `.dmp` file (CrashDumps / minidump).
3. Focus the composer and paste (`Ctrl+V`).

## Expected behavior

The original file is attached as `@path` (same as drag-drop / + Files). Dump files should not be copied through the WebView as base64.

## Actual behavior

`addAttachmentsFromFiles` sees a `File` **without** `path`, then `await f.arrayBuffer()` throws Chromium `NotReadableError`. `formatAttachErrorMessage` classifies it as `other` and appends the English DOMException to `composer.attachPasteFailed`.

## Why

Explorer file copy puts `CF_HDROP` on the clipboard. WebView2 exposes `File` objects with a name, but the blob is often unreadable:

- dumps may still be locked by WER
- dumps are frequently larger than the 40 MiB paste cap
- even when readable, JS `arrayBuffer` + base64 is the wrong path for existing on-disk files

Native clipboard already has the absolute path. Drag-drop uses that; paste does not.

## Suggested fix

1. Host command: read OS clipboard file list (Windows `CF_HDROP`; macOS/Linux file URLs).
2. On paste, if native paths exist, `pathsClassify` + attach — **do not** `arrayBuffer`.
3. Keep `save_temp_attachment` for screenshots / in-memory blobs.
4. Classify `NotReadableError` as a dedicated attach kind; i18n hint: drag in or use + Files. Do not dump Chromium English in the banner.

## App version

Installed SuperGrok (Windows). Error string matches current `formatAttachErrorMessage` (`attachmentsPro.ts`).

## OS

Windows

sutongwuyanzu:
## Summary
- Windows Explorer paste of on-disk files (especially `.dmp`) no longer calls WebView `File.arrayBuffer()`, which throws Chromium `NotReadableError`.
- Host reads the OS clipboard file list (`CF_HDROP` on Windows; `file://` / uri-list elsewhere) and attaches the original `@path`, same as drag-drop.
- Screenshots / in-memory blobs still go through `save_temp_attachment`.
- Unreadable blobs map to `attach.err.unreadable` (15 locales) instead of dumping English `NotReadableError` in the banner.

Fixes #755

## Type of change
- [x] Bug fix

## Test plan
- [x] `clipboardPaste.test.ts` covers OS-file detection, `File.path`, and `NotReadableError`
- [x] `attachmentsPro.test.ts` classifies `NotReadableError` as `unreadable`
- [x] Rust `parse_clipboard_file_list_text` unit test
- [ ] `pnpm typecheck` / `pnpm test` / `cargo test` (CI)
- [ ] Windows: copy a `.dmp` in Explorer, paste into composer → attaches original path
- [ ] Windows: paste a screenshot still saves `attachments/paste/*.png`
- [ ] Drag-drop of the same dump still works (unchanged)

## Checklist
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh + zh-TW + other catalogs)
- [x] CHANGELOG Unreleased
- [x] No secrets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ See `docs/llm-wiki/release.md`.
 - **宠物换成完整 bloub 形变目录**：浮层改用测过的 SVG 引擎（静止、思考、眨眼、通知、警示、六边形、轨道、彗星等），不再用旧的 clip-path 表情。设置可选 8 种休息形体和 16 种休息表情。输入框打字会走目录里的警示（斜感叹号），停打后回到所选休息形体；未读完成会话是显眼的橙红圆点（身体已经很热时改用柠黄），不再用视频里的蓝色。
 
 ### Fixed
+- **Windows paste of Explorer files (.dmp) no longer dies on NotReadableError (#755)**: Composer paste of an on-disk file (crash dumps, locked/large binaries) now reads the OS clipboard path list (`CF_HDROP`) and attaches `@path` like drag-drop. WebView `File.arrayBuffer()` is only used for in-memory blobs (screenshots). Unreadable blobs get a dedicated i18n hint instead of the Chromium English `NotReadableError`.
 - **Windows no longer freezes at end of a long turn (#754)**: A trailing `prompt_complete` after `session/prompt` RPC Ok used to re-arm the end-of-turn handler. The second pass emitted IPC while journal reconcile still held the store lock; the WebView WndProc waited on that lock inside `SendMessage` and the window stopped painting. Duplicate finish is now a no-op, and post-turn UI rehydrate reads App `messages.json` only (Host already merged agent history).
 - **Local session API no longer wedges after the first turn**: A hung ACP handshake used to keep `connect_lock` forever (90s timeout abandoned the waiter without aborting the task). Later `--session-send` / `POST /turns` then waited ~180s and the CLI lied with `app_not_running`. Connect now aborts and bounded-kills pending children; dispatch returns `retry_later` in 15s; health exposes `connectLockBusy`; CLI maps timeouts to `error`; queued external prompts persist on disk and the Host drains them without the main window.
 - **Japanese PR hub and Ukrainian agent command keep `{name}` (#751)**: `ja` `prHub.author` was `作成者` with no author; `uk` `preferredCurrent` shipped `--agent ` with nothing to paste. Catalog tests now fail if any locale drops or renames an `en` placeholder.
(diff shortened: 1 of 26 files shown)


RongleCat (changes requested):
Thanks for the CF_HDROP / `@path` approach — that is the right fix for Explorer `.dmp` paste, and the 15-locale `attach.err.unreadable` catalog is complete.

## Request changes

**`clipboardLooksLikeOsFiles` treats `text/uri-list` as an OS file list.** Firefox (all platforms) and Chromium/Linux put `text/uri-list` on the clipboard when copying a link or a browser image. ComposerEditor now takes that branch *before* the media/text path:

- Paste a copied URL → `onPasteFiles([])` → no native paths → `reportAttachError({ code: "unreadable" })` while the URL still inserts. Previously this was a silent text paste.
- Paste a copied browser image (uri-list + `image/png`, no File, no CF_HDROP) → the OS-files branch returns early and the screenshot/native image fallback never runs.

The new tests only cover `types: ["Files"]` and `["image/png"]`. Please drop the bare `t === "text/uri-list"` check (`Files` / non-image file items already cover Explorer/Finder) and add a regression: `types: ["text/uri-list","text/plain"]` + an http URL must *not* look like OS files.

## Nits (not blocking once the above is fixed)

- Commit `Cargo.lock` with `clipboard-win`.
- `isNotReadableBlobError` is unused in production (classifyAttachError already string-matches).
- `attach.err.unreadable` copy is paste-specific but classify can fire on drop/pick.
- Right-click Paste still reads `navigator.clipboard.readText()` first; if WebView2 surfaces the filename as text, 1b never runs. Worth a Windows check.

CHANGELOG en/zh, no `window.confirm`, no App.tsx state growth. Happy to re-review after the uri-list gate is tightened.

RongleCat (approved):
uri-list gate is tightened: `clipboardLooksLikeOsFiles` no longer treats bare `text/uri-list` as an OS file list. Copied http URLs stay on the text path; Explorer `Files` / non-image file items still attach via `CF_HDROP`. Regression test covers `types: ["text/uri-list","text/plain"]`.

CHANGELOG stacks on current Unreleased (does not clobber #754). Thanks @sutongwuyanzu.

RongleCat:
Thanks @sutongwuyanzu — Explorer `.dmp` paste via CF_HDROP / `@path` is the right fix. Squash-merged (uri-list gate tightened so copied URLs stay text). Fixes #755.

PR #759, opened 2026-08-20
fix(session): stop local session API stalling after the first turn

RongleCat:
## Summary
- Hung ACP handshakes no longer pin `connect_lock` forever. The 90s wall-clock now lives **inside** the connect sibling, so dropping a 15s HTTP waiter only detaches; the sibling still times out, `fail_stale_connecting`, and bounded-kills pending children.
- `POST /turns` returns `503 retry_later` within 15s. CLI maps transport timeout/5xx to `error` (exit 1), not `app_not_running`. `/v1/health` reports `connectLockBusy`.
- External queued prompts persist on disk. Host drains them; the main window only displays `source: external`. CLI HTTP timeout is 20s so it can see `retry_later`.
- Lock-holder diagnostics escalate after ~6 minutes of idle-recycle skips.

Windows users hitting this after `--session-send` / Codex dispatch: still set system `HTTPS_PROXY` and fully quit the tray so children inherit the proxy. After this change the host recovers in ≤90s instead of needing a restart.

## Type of change
- [x] Bug fix

## Test plan
- [x] `cargo test --lib connect_timeout` (5)
- [x] `cargo test --lib session_api::` (14)
- [x] `vitest run src/lib/sendQueue.test.ts` (35)
- [ ] Windows: disconnect proxy, `--session-send` twice; health stays fast; second send is `retry_later` then recovers; no extra idle `grok.exe`

## Checklist
- [x] User-facing API strings are machine-facing JSON (`retry_later`, health field); docs in `session-api.md`
- [x] CHANGELOG Unreleased
- [x] No secrets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ See `docs/llm-wiki/release.md`.
 - **宠物换成完整 bloub 形变目录**：浮层改用测过的 SVG 引擎（静止、思考、眨眼、通知、警示、六边形、轨道、彗星等），不再用旧的 clip-path 表情。设置可选 8 种休息形体和 16 种休息表情。输入框打字会走目录里的警示（斜感叹号），停打后回到所选休息形体；未读完成会话是显眼的橙红圆点（身体已经很热时改用柠黄），不再用视频里的蓝色。
 
 ### Fixed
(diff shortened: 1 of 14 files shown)


RongleCat:
Squash-merged to main as 808549bd. Remote `fix/session-api-stall` deleted. Pre-merge pi review: no blockers (B1 still fixed, CHANGELOG Unreleased appended not clobbered, no Fixes/Closes to close).

PR #772, opened 2026-08-21
fix(chat): stop bounce-up when images and PDF previews decode

### LINKED ISSUE #771: [Bug]: Long chats bounce upward while images/PDF previews load
## Describe the bug

On a long chat that stays pinned at the bottom, the transcript keeps jumping up and flashing while the assistant turn is still drawing. Worst when that turn dumps many screenshots or PDF page previews (Grok reading PDFs / pasting screenshots in one go). Each file decode grows the last rows, and the scroller snaps to the new max over and over.

This is not #703 / #714 / #760. Those covered trackpad unpin, send/settle flash, and double force-stick. This still happens on current `main`.

Root cause: two systems write `scrollTop` to max on every content-height jump, and they do not share an ignore flag.

1. `useChatMessageVirtualizer.commitRowHeight` writes `scrollTop = max` on each row ResizeObserver (image / PDF decode).
2. The virtual window then rebuilds; a layout effect writes `scrollTop = max` again.
3. `useStickToBottom`'s content ResizeObserver calls `followIfPinned()` in the same stretch.

A row of screenshots is one snap per file, not one snap for the turn. The viewport sits off the new max for a moment, then the writers yank it down, so the messages fly up. Repeat until decode finishes. Same class of bug as other chat-window shivers: stick-to-bottom and the virtual list both own the scrollbar.

## Steps to reproduce

1. Windows Grok App, long session (enough rows to virtualize, 16+).
2. Stay pinned at the bottom (do not scroll up).
3. Ask something that returns several screenshots or PDF page images in one assistant turn (tool output / file reads).
4. Watch the transcript while those images decode.

## Expected behavior

The list follows the new tail once after the images settle. No repeated bounce-up or flash.

## App version

0.2.24 / current main after #760.

## OS

Windows (reproduced in a live Grok App session). Same path on any OS with a virtualized transcript.

---

**中文**

长对话贴在底部时，助手一回合里甩出一堆截图或 PDF 页预览，会话会反复往上弹、闪一下。不是 #703 / #714 / #760 那几类（轻滑被拽回、发送双吸底、发送/收尾整页闪）。

原因：贴底和虚拟列表各写一遍 `scrollTop`。每张图解码，行高 ResizeObserver 先吸一次底，窗口重算 layout 再吸一次，stick 的内容观察再吸一次。一串图就是弹一串。对话窗口里同类问题的源头也是这两套滚动逻辑抢同一条滚动条。

**复现**

1. Windows Grok App，长会话（够虚拟列表，16 条以上）。
2. 人贴在底部，不要往上翻。
3. 让助手一回合读 PDF / 贴多张截图。
4. 图解码的那几秒看会话：会一下一下往上弹。

期望：图落定后吸一次底，不要弹一串。

erict16:
## Summary
Long chats pinned at the bottom bounce upward while screenshots / PDF page previews decode. Stick-to-bottom and the virtual list both wrote `scrollTop = max` on every row height jump.

This PR:
- Stops per-row pinned snaps in `commitRowHeight` (one snap after the coalesced window layout)
- Trailing-coalesces stick follows for media-sized height jumps (24px+) so a decode storm is one follow, not one per file

Fixes #771.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] vitest: `stickToBottom` + `chatVirtualList` (84 tests)
- [ ] full `tsc -b` (pre-existing `@testing-library/*` errors in workbench-modals tests, not this change)
- [ ] I ran `cargo test` in `src-tauri` (or `cargo check` for docs-only)
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh) — n/a, no copy

### DIFF
--- a/src/hooks/useChatMessageVirtualizer.ts
+++ b/src/hooks/useChatMessageVirtualizer.ts
@@ -11,6 +11,8 @@
  * - Per-row ResizeObserver so image/video decode updates height cache (callback
  *   refs alone only fire on mount).
  * - Debounced recompute so measure storms cannot oscillate the window.
+ * - Pinned: no per-row scrollTop snap. Image/PDF decode used to snap on
+ *   every commit, then the window layout snapped again (bounce-up).
  *
  * Long-session perf:
  * - rAF-coalesce scroll recomputes (one window update per frame while flinging).
@@ -35,6 +37,7 @@ import {
   resolveChatOverscanPx,
   scrollTopAfterHeightChange,
   shouldCommitRowHeight,
+  shouldWriteScrollOnRowCommit,
   shouldVirtualizeChat,
   type ChatVirtualWindow,
 } from "@/lib/chatVirtualList";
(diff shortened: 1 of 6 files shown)


RongleCat:
Landed via #779 (same commit plus the slow scroll-up fix). Thanks @erict16.

PR #774, opened 2026-08-21
fix(windows): use OS maximize for caption button, not work-area fill

zhangxaochen:
## Summary
- Windows caption **maximize** does not OS-maximize. The glyph stays a square. After restore, dragging the top edge pans the **page inside the frame**.
- `00dd8ef5` (#628) replaced `toggleMaximize()` with `toggleMaximizeReliable()`. The Linux work-area `setSize` fill runs when `isMaximized()` still lags ~40ms. That path was not gated to Linux, so Windows `setSize` cancels a settling OS maximize. Titlebar double-click still hits native caption maximize, so it looks better.
- Windows/mac now wait for the OS flag and never fake-fill. Caption switches to the restore glyph. WebView2 `visualViewport` pan is pinned. Linux work-area fill is unchanged. Frameless HWND also keeps `WS_MAXIMIZEBOX`.

Fixes #773

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Test plan
- [x] `pnpm exec vitest run src/lib/windowChrome.test.ts src/lib/window-config.test.ts` (15 passed)
- [x] `pnpm exec eslint` on touched TS (clean)
- [ ] `cargo test --lib win_shell::tests` — harness compiled; run aborted with `STATUS_ENTRYPOINT_NOT_FOUND` on this host (same environment issue as nearby Windows Host tests)
- [ ] CI green
- [ ] Windows: caption maximize is real OS maximize + restore glyph; restore; drag the top edge — UI stays pinned in the frame
- [ ] Titlebar double-click still toggles maximize
- [ ] Linux: compositor no-op still fills the work area

## Checklist
- [x] I ran `pnpm typecheck` (no errors in touched files) and targeted vitest
- [ ] `cargo test` in `src-tauri` (see test plan)
- [x] No new i18n keys (reuses `window.maximize` / `window.restore`)
- [x] No `window.confirm` / `prompt` / `alert`
- [x] CHANGELOG Unreleased (en + zh)
- [x] No secrets

Note: `pi -p` is not installed on this Windows host (no `pi` on PATH). Per repo policy this is recorded rather than skipped silently.

Issue #773, opened 2026-08-21
[Bug]: Windows maximize button does not maximize; top-edge drag after restore pans the UI


zhangxaochen:
### Describe the bug

On Windows, the frameless caption **maximize** button does not enter a real OS maximize. The glyph stays a single square (never the restore / overlapping-squares icon). After clicking it again (restore), dragging down from the top edge of the window makes the **page inside the frame** walk downward; the HWND itself does not run away on the desktop.

Titlebar double-click is closer to correct because it still hits native caption maximize.

Root cause: `00dd8ef5` (#628, Linux window chrome) replaced `toggleMaximize()` with `toggleMaximizeReliable()`. That helper waits ~40ms and, if `isMaximized()` is still false, `setSize`-fills the monitor work area. The fallback was meant for GTK/Wayland no-ops but was **not gated to Linux**. On Windows the follow-up `setSize` cancels a maximize that was still settling. Fake geometry also leaves WebView2 able to pan `visualViewport` when the user resizes from the top.

This is not in the last 15 local commits (Win+Right / Alt+Tab / pet / shortcuts). Caption glyph never switching to restore is older (`IconMaximize` was always rendered); the false maximize + inner pan started at #628.

### Steps to reproduce

1. Windows, Grok App, normal (not maximized) window
2. Click the top-right maximize square (not titlebar double-click)
3. Observe: window is not OS-maximized; icon still a square
4. Click the same button again
5. Grab the top edge and drag downward

### Expected behavior

- Maximize button calls OS maximize, fills the work area, and the glyph switches to restore
- Restore returns the previous size
- Dragging the top edge resizes the window; the UI stays pinned inside the frame

### App version

0.2.24 (current main)

### OS

Windows

PR #784, opened 2026-08-21
fix(windows): defer caption maximize and restore titlebar move

### LINKED ISSUE #783: [Bug]: Windows maximize button flashes then restores; titlebar drag grows the window downward
### Describe the bug

Follow-up to #773 / #774. Titlebar **double-click** maximizes and the restore glyph is correct. Clicking the caption **maximize button** still flashes maximized and immediately restores.

Separately, when the window is **not** maximized, dragging the titlebar **up** does not move the window: the bottom edge stays put (or grows), as if the titlebar were a north resize handle. You cannot lift the window.

Cause:
1. `maximize()` runs during the same click while the pointer is still down. Windows treats that as Aero drag-to-restore.
2. #774 pinned `visualViewport` with a `body` CSS transform. That breaks `-webkit-app-region: drag`, so titlebar moves become HTTOP.

### Steps to reproduce

1. Windows, Grok App, not maximized
2. Click the top-right maximize square (not titlebar double-click)
3. Observe: flash then original size
4. Drag the titlebar (middle of the 40px band) upward

### Expected behavior

- Caption maximize stays maximized; glyph is restore
- Dragging the titlebar moves the whole window; the bottom edge comes with it
- The very top few pixels may still north-resize

### App version

main after #774 (0.2.24 + Unreleased)

### OS

Windows

zhangxaochen:
## Summary
- Follow-up to #773 / #774. Titlebar double-click maximizes correctly; **clicking the caption maximize button** flashes then restores.
- `maximize()` ran during the same click while the pointer was still down. Windows treats that as Aero **drag-to-restore**.
- #774 pinned `visualViewport` with a `body` transform. That breaks `-webkit-app-region: drag`, so dragging the titlebar **up** north-resizes (bottom edge never lifts).
- Defer caption `maximize()` 32ms past mouse-up. Drop the transform pin. 8px top no-drag strip keeps native HTTOP.

Fixes #783

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Test plan
- [x] `pnpm exec vitest run src/lib/windowChrome.test.ts` (5 passed)
- [ ] CI green
- [ ] Windows: click caption maximize — stays maximized, restore glyph
- [ ] Windows: not maximized, drag the **middle** of the titlebar up — whole window moves
- [ ] Top few pixels still north-resize
- [ ] Titlebar double-click still toggles maximize
- [ ] Linux work-area fill path unchanged

## Checklist
- [x] Targeted vitest
- [x] No new i18n keys
- [x] CHANGELOG Unreleased (en + zh), separate from #773
- [x] No secrets
- [x] One commit on `origin/main` (does not rewrite #774)

Note: `pi -p` is not installed on this Windows host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ See `docs/llm-wiki/release.md`.
 - **仓库 About 与 README 挂上官网**：GitHub 仓库网站、`package.json` `homepage` 与各语言 README 现指向 [https://grok-app.com/](https://grok-app.com/)。
 
 ### Fixed
-- **Windows maximize button is a real maximize (#773)**: The caption square used a Linux work-area `setSize` fill when `isMaximized()` lagged ~40ms. That cancelled the OS maximize, left the glyph as a square, and after restore, dragging the top edge panned the WebView inside the frame. Windows now waits for the OS flag and never fakes geometry; the button switches to the restore glyph; visualViewport pan is pinned.
+- **Windows caption maximize stays maximized; titlebar drag moves the window (#783)**: Follow-up to #773/#774. The button defers `maximize()` until after mouse-up so Aero does not drag-to-restore (flash). The `body` visualViewport transform pin is gone — it broke `-webkit-app-region: drag`, so pulling the titlebar up north-resized (bottom never lifted). An 8px top no-drag strip keeps native HTTOP.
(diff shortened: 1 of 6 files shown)


RongleCat:
Merged, thanks @zhangxaochen — Fixes #783. Caption maximize defers past mouse-up; titlebar drag no longer north-resizes.

PR #809, opened 2026-08-21
perf(chat): keep history message identity when weaving tools

### LINKED ISSUE #800: [Perf]: keep history message identity when weaving tools
### Problem / motivation
`weaveToolsIntoAssistantSegments` clones every transcript row on each stream notify, so memoized chat rows fail `a.m !== b.m` and `wovenMessages` reference checks.

### Proposed solution
Slice instead of map-clone; only replace assistants that actually change. Drop the whole-array `wovenMessages` identity check on the row memo.

### Additional context
Wave 2 B1.

zhangxaochen:
## Summary

`weaveToolsIntoAssistantSegments` cloned every chat row on each stream notify, so memoized transcript rows failed `a.m !== b.m`. It now slices the array and only replaces assistants it actually rewrites. Row memo no longer busts on a new `wovenMessages` array.

Fixes #800

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Test plan

- [x] `pnpm exec vitest run src/lib/session.test.ts -t weaveToolsIntoAssistantSegments` (6 passed, including history identity)
- [ ] Stream a long tool-heavy answer: older bubbles should not remount
- [ ] CI green

## Checklist

- [x] Targeted vitest
- [x] No new i18n keys
- [x] CHANGELOG Unreleased (en + zh)
- [x] No secrets

## Notes

- `pi -p` is not on this Windows PATH. Do not treat as pi-reviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ See `docs/llm-wiki/release.md`.
 - **快捷键帮助（Ctrl+/）**：面板可按名称 / id / 组合键筛选，按设置页同样的分组列出；补上缩放、换行、历史提示、打字聚焦；列表可滚动，不再裁掉末尾几项。
 
 ### Changed
+- **Weaving tools no longer clones every chat row**: History message object identity is kept so memoized transcript rows survive stream ticks. Row memo no longer busts on a new `wovenMessages` array.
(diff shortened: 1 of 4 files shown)


RongleCat:
Merged, thanks @zhangxaochen

PR #832, opened 2026-08-22
perf(files): window long resource trees

### LINKED ISSUE #831: [Perf]: files tree should not mount every expanded row
### Problem / motivation
The files tree recursively maps every expanded node into the DOM. Expanding a large folder (or a filter that opens many dirs) mounts one button per entry.

### Proposed solution
Flatten the visible rows and window them. Short trees stay a full list. At 32+ visible rows, only the viewport is mounted. Row height stays 28px.

### Additional context
`FilesWorkspace` and `ResourceViewer` both have recursive `renderTree`. Independent of markdown preview and chat-find.

zhangxaochen:
## Summary

Side-pane / resource files trees keep short folders as a full list. At 32+ visible rows, only the viewport is mounted (28px rows), instead of recursively mapping every expanded entry.

Fixes #831

## Type of change

- [x] Refactor / chore
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

## Test plan

- [x] `pnpm exec vitest run src/lib/resourceTree.test.ts` (14 passed)
- [ ] Open a small project tree: expand/collapse and open a file unchanged
- [ ] Expand a large folder (32+ visible rows): scroll stays responsive
- [ ] Filter still shows matching files and ancestor dirs
- [ ] Active file still highlights; opening a file from the tree still works
- [ ] CI green

## Checklist

- [x] User-facing strings go through `t()` / `createT`
- [x] CHANGELOG Unreleased (en + zh)
- [x] No secrets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ See `docs/llm-wiki/release.md`.
 - **快捷键帮助（Ctrl+/）**：面板可按名称 / id / 组合键筛选，按设置页同样的分组列出；补上缩放、换行、历史提示、打字聚焦；列表可滚动，不再裁掉末尾几项。
 
 ### Changed
+- **Files tree windows long listings**: Side-pane / resource trees keep short folders as a full list. At 32+ visible rows, only the viewport is mounted (28px rows), instead of recursively mapping every expanded entry.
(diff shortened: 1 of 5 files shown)


RongleCat:
Merged, thanks @zhangxaochen — long files trees window the viewport instead of mounting every expanded row.

PR #839, opened 2026-08-22
feat(settings): pick UI and terminal fonts from installed families

Dmao233:
Replace the free-text Appearance font fields with a searchable dropdown
of local font families. Host lists Core Text / GDI / fontconfig names;
empty still means the default UI sans stack or the built-in Nerd Font.

Also use a white inverted dock icon for `pnpm dev` so the debug app is
distinct from the installed release, and ignore dev heartbeat files so
Vite does not full-reload the splash.


RongleCat:
Thanks @Dmao233 — squash-merged as **#839**.

Settings → Appearance now picks installed UI / terminal families from a searchable dropdown. Debug `pnpm dev` uses the white inverted dock icon so it stays distinct from the installed Grok app.

PR #847, opened 2026-08-22
fix(settings): keep overlay full-window when wallpaper is on

### LINKED ISSUE #846: [Bug]: Settings does not fill the window when a wallpaper is set
### Describe the bug

Open Settings with a wallpaper on. Short pages (Pet, Archived chats, and likely About / Keyboard) do not cover the window. The overlay stops where the content stops. The wallpaper shows through a big empty strip at the bottom, and the desktop pet can sit in that gap.

Longer pages (General, Appearance) hide it because their content is tall enough to fill the frame. Switch to Pet or Archived chats and the hole comes back.

This is not a missing background on those two tabs. The settings overlay itself is no longer `position: absolute; inset: 0`. With wallpaper on, this rule in `src/styles/skins.css` lifts every direct child of `.app-shell` except the titlebar and the wallpaper layer:

```css
html[data-wallpaper="1"] .app-shell
  > *:not(.window-controls):not(.app-wallpaper-media) {
  position: relative;
  z-index: 2;
}
```

That was meant to keep the workbench above the wallpaper. It also hits `.app-settings-stage`. Relative positioning drops `inset: 0`, so the stage becomes a normal column flex item and sizes to its content. The still-mounted workbench keeps `flex: 1` and `opacity: 0` underneath, which is why you see wallpaper (and the pet) in the leftover space.

Same family of bug as the window-controls carve-out already in that rule: `position: relative` on a child that must stay `fixed` / `absolute`.

### Steps to reproduce

1. macOS Grok App, wallpaper set (Appearance → wallpaper). Window not necessarily maximized.
2. Open Settings (account menu, or the usual shortcut).
3. Open **Archived chats**. Look at the bottom of the window.
4. Open **Pet**. Same gap.
5. Optional: open General or Appearance. Those often look full because the page is tall. Switch back to Pet / Archived chats and the strip returns.

### Expected behavior

Settings should cover the whole window on every tab, wallpaper or not. Nav + content stretch to the bottom. Wallpaper can still show *through* the frosted panes. It should not show as a leftover hole under a short page.

### App version

0.2.24 (installed Grok on macOS). Still present on current `main` (`src/styles/skins.css` wallpaper lift vs `.app-settings-stage { position: absolute; inset: 0 }` in `settings.part1.css`).

n/a (UI layout)

### OS

macOS. The CSS is not Mac-only; any platform with a wallpaper should hit the same overlay shrink.

### Logs / screenshots

Reproduced on the live app:

- Settings → Archived chats: overlay ends above the window bottom; wallpaper fills the rest.
- Settings → Pet: same cut. Desktop pet sits in the uncovered strip.

Fix direction: do not force `position: relative` on `.app-settings-stage` (and probably `.setup-gate`). The wallpaper lift only needs `.workbench`.

erict16:
## Summary

Open Settings with a wallpaper, then go to Pet or Archived chats. The page does not fill the window. Overlay stops where the content stops. Wallpaper (and the desktop pet) sit in the leftover strip at the bottom.

Longer tabs hide it because they are tall enough. The hole is not a missing fill on those two pages. The settings overlay itself was no longer `position: absolute; inset: 0`.

Wallpaper CSS was lifting **every** direct child of `.app-shell` except the titlebar and the media layer:

```css
html[data-wallpaper="1"] .app-shell
  > *:not(.window-controls):not(.window-edge-n):not(.app-wallpaper-media) {
  position: relative;
  z-index: 2;
}
```

That was meant to keep the workbench above the wallpaper. It also hit `.app-settings-stage`. Relative positioning drops `inset: 0`, so the stage becomes a column flex item and sizes to its content. The still-mounted workbench keeps `flex: 1` and `opacity: 0` underneath.

Same class of bug as the window-controls carve-out already in that rule.

This PR only lifts `.workbench`. Settings stays `absolute; inset: 0` and covers the window on every tab.

Fixes #846.

## How we tested it

- Guard test `src/lib/settingsOverlayFill.guard.test.ts`: lift selector is `.workbench` only; settings stage still `position: absolute; inset: 0`; no `position: relative` on the stage.
- `pnpm exec vitest run src/lib/settingsOverlayFill.guard.test.ts` — 3 passed.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] vitest: `settingsOverlayFill.guard.test.ts` (3 passed)
- [ ] `pnpm typecheck` / full `pnpm test` — CSS + guard only, no TS surface
- [ ] `cargo test` — no Rust change
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh) — n/a, no copy
- [x] Docs / `docs/llm-wiki` updated if behavior changed — n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,7 @@ See `docs/llm-wiki/release.md`.
 - **仓库 About 与 README 挂上官网**：GitHub 仓库网站、`package.json` `homepage` 与各语言 README 现指向 [https://grok-app.com](https://grok-app.com)（无末尾斜杠）。
 
 ### Fixed
+- **Settings covers the window with wallpaper on (#846)**: Wallpaper chrome lift no longer forces `position: relative` on the settings overlay. Short tabs (Pet, Archived chats) were shrinking to content height, so wallpaper (and the desktop pet) showed in the leftover strip.
 - **Agent questionnaire no longer freezes the workbench (#844)**: Answering or dismissing `_x.ai/ask_user_question` used to wait on the Host IPC with every control disabled. The modal now closes first and restores only if a failed accept is still the live request. Portaled overlays also opt out of the window-drag region so macOS titlebar drag cannot swallow clicks.
 - **Windows titlebar drag-up no longer grows height (#786)**: Follow-up to #783/#784. Caption maximize stayed up, but dragging the titlebar still stretched the frame. Windows skips JS `start_dragging` (`data-tauri-drag-region="false"`) and keeps compositor caption drag. Host `set_min_size` no longer runs on every `Moved` (tao re-applies inner size and double-counts the shadow offset); it skips while maximized, while the pointer is down, and when the min is unchanged.
 - **Desktop pet stays the chosen body, celebrates once, and no longer stalls on Windows**: Typing and in-progress tools no longer morph the mark into the catalog `!` or orbit/comet ribbons — the rest shape stays, with an attentive/curious face. Colorful belts fire only when the last live turn becomes unread-ready, then a corner pastille. Overlay paint throttles after idle, pauses while hidden, and cursor look events stay on the pet window (no duplicate still-cursor wakeups) so a long-running Windows overlay no longer freezes until restart.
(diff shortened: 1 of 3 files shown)


erict16:
Live-app screenshots for #846.

**Settings → Archived chats**

Settings Archived chats not filling the window

**Settings → Pet**

Settings Pet not filling the window

RongleCat:
Reviewed against the maintainer checklist.

Root cause matches #846: wallpaper chrome lift was forcing `position: relative` on `.app-settings-stage`, which dropped `inset: 0` so short tabs sized to content. Lifting only `.workbench` is the right carve-out (also leaves `.setup-gate` alone). Guard test locks the selector.

No in-app copy, no native dialogs, no secrets. CI green.

Squash-merging. Follow-up #848 still needs a rebase of its `skins.css` wallpaper lift onto this narrower selector.

RongleCat:
Thanks @erict16 — landed on main as #847 (Fixes #846).

PR #848, opened 2026-08-22
fix(settings): switch views atomically without blank frames

Dmao233:
## Summary

- Replace the delayed Settings presence with an atomic view switch while keeping the workbench painted and inert underneath.
- Close account-menu portals and cover native child webviews before Settings navigation, including hash/deep-link entry.
- Align Settings/sidebar stacking, frost, width, and seam behavior, with regression tests for blank-frame and flash paths.

## Root cause

The Settings stage became visually hidden immediately on close but remained mounted for 320 ms. In WKWebView, repainting the underlying workbench could be deferred until that stage unmounted, which exposed wallpaper-only frames. Lazy first-load and overlay/stacking paths added one-frame flashes around the same transition.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Validation

- User verified the native macOS transition in both directions.
- `pnpm test` — 490 files / 6481 tests passed.
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build:ui`
- `cargo test --locked -- --test-threads=1` — 1480 passed / 1 ignored.
- `pnpm tauri build --debug --bundles app --config src-tauri/tauri.dev.conf.json`
- Chromium regression: 5 flows / 80 commit-frame samples, including wallpaper and hash navigation.
- Final integrated Pi review: PASS.

The default parallel Cargo run hit one shared temporary-directory rename race; that test passed in isolation and the full serial suite passed.

## Checklist

- [x] I ran `cargo test` in `src-tauri` (or `cargo check` for docs-only)
- [x] No new user-facing strings were added
- [x] Behavior docs and changelog were updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ See `docs/llm-wiki/release.md`.
 - **快捷键帮助（Ctrl+/）**：面板可按名称 / id / 组合键筛选，按设置页同样的分组列出；补上缩放、换行、历史提示、打字聚焦；列表可滚动，不再裁掉末尾几项。
 
 ### Changed
+- **macOS workbench sidebar uses the same frost as Settings**: Left rail (sessions + Settings nav) applies `--sidebar-blur` backdrop-filter instead of relying on native vibrancy alone. Settings is available before first navigation and swaps atomically with the workbench; the account portal closes in the same commit. Settings nav width follows the workbench rail. The glass meets the chat column with no opaque black seam. Returning keeps the workbench painted instead of hiding Settings for 320ms and leaving only the wallpaper in WKWebView.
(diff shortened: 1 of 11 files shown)


RongleCat:
Rebased onto `main` after #847.

Kept the #846 wallpaper lift (`html[data-wallpaper="1"] .app-shell > .workbench` only). Restoring the old `> *:not(...)` selector would shrink short Settings tabs again. Settings still stacks at `z-index: 20` for the atomic swap. Guard test updated.

Maintainer checklist: no new copy, no native dialogs, no secrets. Local vitest for overlay fill / openPresence / paneSplitMotion passed.

RongleCat:
Thanks @Dmao233 — landed on main as #848.

PR #829, opened 2026-08-22
ci(rust): clear Linux/Windows clippy residue, gate all legs at -D warnings

shiaho777:
Fixes #828

## Status: ready for review — machine gates green

- **CI: all four jobs green** under `cargo clippy --all-targets -- -D warnings` (Linux / Windows / macOS legs + frontend).
- Local: clippy `-D warnings` green, `cargo test` 1423 passed, `cargo fmt --check` clean.
- **pi 审核状态**：AGENTS.md §8 的强制 `pi -p` 审核因服务持续 503 暂未完成（2026-08-22，13+ 次重试）。repo owner 指示先行转正；定时任务会在 pi 恢复后自动补跑完整审核、把记录落到 `docs/plans/2026-08-22-pr829-pi-review-fix.md` 并在本 PR 追加结论。

## Problem

701bf7b9 restored zero clippy warnings on a macOS host, but CI keeps clippy advisory (`-W clippy::all`). The latest green main run (32540900892) still logs per leg: **Linux ~18, Windows ~31, macOS 3** — platform-cfg dead code, unused imports on non-native legs, block-tail returns, and newer-toolchain lints an older local toolchain doesn't see. The debt regressed to 98 warnings on macOS alone before the last sweep; `-W` cannot stop that.

## Approach

Same idioms as the existing sweeps:

- **Platform-gate** prod code only one OS wires up: editors `sips`/base64/`icon_cache_dir` + schedules launchd helpers (macOS), `APP_BUNDLE_ID` (not linux), proxy PAC parser + `SystemSocks`/`SystemPac` variants (not linux), apple languages tag (macOS).
- **Drop returns** that are provably the block tail of their single compiled cfg branch; early returns inside `if`/`if let` arms are load-bearing and kept. Windows leg runs `cargo test`, so any mistake there surfaces in this PR's CI.
- **Remove imports dead on their own platform** — notably `std::os::windows::process::CommandExt`: tokio's `creation_flags` is an inherent method, so the trait import is unused even on Windows.
- **Newer-toolchain lints**: `as_chunks::<N>()` for constant chunk sizes (clippy 1.98 `chunks_exact_to_as_chunks`), `&Path` params, `upper_case_acronyms` allow for the Win32 `BOOL` mirror, `result_large_err` allow on `await_stdio_or_cancel` (matches the existing `prepare_send` precedent).
- **Flip the gate**: `cargo clippy --all-targets -- -D warnings` on all three legs.

## Test plan

[...]

PR #851, opened 2026-08-22
fix(remote-im): authenticate WeCom webhook and make the sender ACL fail-closed

KronixDev:
## Motivation

While reviewing the Remote IM subsystem we found two related gaps that together let an attacker on the LAN drive the local agent through forged IM messages:

**1. WeCom webhook accepts unauthenticated messages and binds `0.0.0.0`**

`src-tauri/src/remote_im/channels/wecom.rs` (webhook mode):

- POSTs were parsed as JSON with **no `msg_signature` verification** — anyone who can reach the port can forge an `IncomingMessage` with an arbitrary `FromUserName`.
- The GET verify endpoint echoed any `echostr` back without checking anything.
- The listener bound `[0,0,0,0]:port` unconditionally (every other channel binds loopback unless explicitly opted in; LINE documents this contract).

Combined with **2**, this was a remote-to-local agent-execution path from any LAN host.

**2. The sender ACL was open by default in the backend**

`outbound.rs::allow_from_list` parsed a *missing* `allowFrom` as `"*"` (allow everyone), so every gate (`engine.rs` text path + card actions, telegram connector) treated unknown senders as allowed when the entry was absent.

Notably, the rest of the product already treats an empty allow-list as fail-closed:

- The Settings UI refuses to enable a channel without an explicit allow-from entry (`settings.remoteIm.err.allowFromRequired`, translated in all 16 locales).
- `allowFromHelp` strings tell users to prefer explicit ids over `*`.
- `outbound.rs::allow_from_blocks_enable` existed precisely to enforce this but was dead code — never called outside tests.

So the UI said "fail-closed required" while the backend (the actual security boundary) stayed open for hand-edited configs or instances saved before that rule existed.

### Commit 1 — WeCom webhook hardening

- Verify the official WeCom callback signature: `SHA1` over the lexicographically sorted `[token, timestamp, nonce, payload]`, hex-encoded, compared constant-time — for both GET verify and POST callbacks. Query params are percent-decoded before hashing.
- Echo `echostr` only after successful verification (previously reflected any value).
- Require the `callback_token` secret before starting the webhook (fail closed). This secret was already collected as **required** by the Settings UI in webhook mode (`wecomRequiredSecretKeys`) but never checked by the backend — this finishes that wiring.
- Bind `127.0.0.1` by default; opt-in LAN bind via `allow_external=true`, same contract as the LINE channel.
- Exact-path routing on the decoded request target (was a substring check on the whole request).
- New dep: `sha1 = "0.10"` (same RustCrypto family as the existing `sha2`/`hmac`; `hex` was already present).

### Commit 2 — ACL fail-closed at runtime

- `allow_from_list`: missing/empty `allowFrom` now yields a deny-all list; `"*"` remains the documented explicit opt-in.
- `start_runtime` refuses to start an instance without an explicit allow-from entry and surfaces the reason via `set_instance_last_error`, so Settings shows the error instead of a green "connected".
- Semantics documented on both functions.

### Commit 3 — CI hygiene

- Clears two new clippy 1.94 `nonminimal_bool` warnings so `cargo clippy --lib -- -D warnings` stays green (same spirit as #829).

## User impact / migration

- WeCom webhook users must fill the `callback_token` they already see in Settings (it was a required field); without it the channel refuses to start with an explicit error.
- WeCom webhook binds loopback by default. If your callback needs LAN reachability (e.g. you terminate TLS on another box), set `allow_external: true`.
- Channels configured without any allow-from entry will refuse to start with an actionable error instead of accepting everyone. Users who truly want open access set `allowFrom: "*"` explicitly.
- WebSocket mode (the default for WeCom) is unchanged.

## Verification

```bash
cd src-tauri
cargo test --lib                   # 1489 passed, 0 failed
cargo clippy --lib -- -D warnings  # clean
cargo fmt --check                  # clean
```

New/updated Rust tests:

- Signature accept/reject matrix (wrong payload/token/timestamp, missing/empty sig), part-order independence
- Constant-time compare basics
- Percent-decoding incl. truncated escapes
- Query-param extraction (prefix keys must not match)
- Request splitting with and without body
- `wecom_webhook_without_callback_token_fails_closed` — refuses start, nothing listening
- `wecom_webhook_start_binds_port` — updated with token, still binds
- ACL deny-by-default matrix, snake_case alias, wildcard opt-in, comma-list exactness, enable-gate cases

Regression reasoning: only three places construct instances with empty ACLs in tests (`protocol_start_tests`, `weixin_flow_tests`, telegram/weixin unit tests) — all exercise transport-level code below the engine gate, unaffected. `engine.rs`'s own test already sets `allowFrom: "*"` explicitly.

## Notes for maintainers

- The shared-token fallback header (`x-grok-wecom-token`) exists because plain-JSON callback deployments cannot compute `msg_signature`; deployments using it should keep the token long and treat it like a password. Happy to drop it if you would rather be strictly official-scheme-only.
- Follow-up worth considering (not in this PR): WS mode puts `bot_secret` in the connect URL query string; that is mandated by the WeCom endpoint shape, but logging hygiene around it could be audited once more.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1874,6 +1874,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
+ "sha1",
  "sha2",
  "tauri",
  "tauri-build",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -49,6 +49,7 @@ chrono = { version = "0.4", features = ["serde", "clock"] }
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "stream", "multipart", "system-proxy", "blocking", "socks"] }
 futures-util = "0.3"
 base64 = "0.22"
+sha1 = "0.10"
 sha2 = "0.10"
 hmac = "0.12"
 hex = "0.4"
--- a/src-tauri/src/os_theme.rs
+++ b/src-tauri/src/os_theme.rs
@@ -14,7 +14,7 @@ pub const OS_THEME_CHANGED_EVENT: &str = "os-theme://changed";
 /// `AppsUseLightTheme` DWORD: `0` = dark apps, `1` = light. Missing → dark.
 #[cfg_attr(not(test), allow(dead_code))]
 pub fn apps_prefer_dark_from_dword(apps: Option<u32>) -> bool {
-    !apps.is_some_and(|v| v != 0)
+    apps.is_none_or(|v| v == 0)
 }
 
 /// Best-effort OS dark/light probe (no extra deps).
--- a/src-tauri/src/remote_im/channels/wecom.rs
+++ b/src-tauri/src/remote_im/channels/wecom.rs
@@ -146,6 +146,92 @@ fn parse_ws_msg(inst: &ChannelInstance, v: &Value) -> Option<IncomingMessage> {
     })
 }
 
+/// Default local webhook port for WeCom callback mode.
+pub const DEFAULT_WEBHOOK_PORT: u16 = 8081;
+
+/// Shared-token fallback header when `msg_signature` is unavailable
+/// (plain-JSON deployments that cannot compute the WeCom signature).
+pub const SHARED_TOKEN_HEADER: &str = "x-grok-wecom-token";
+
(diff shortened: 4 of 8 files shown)


RongleCat:
Merged — thanks @KronixDev!

PR #906, opened 2026-08-24
fix: do not snapshot() while holding inner (connect no-op deadlock)

### LINKED ISSUE #905: Session API second turn deadlocks: snapshot() while holding inner (parking_lot)
## Describe the bug

`SessionManager::connect_inner` already-live no-op deadlocks on the second Session API turn.

`parking_lot::Mutex` is **not reentrant**. The already-live path locks `self.inner` and then calls `self.snapshot()`, which locks `self.inner` again. The thread never returns, so the tokio `connect_lock` guard never drops. The 90s wall-clock timeout cannot fire because the thread is blocked on a sync mutex, not `.await`.

Health then reports `connectLockBusy: true` forever (`held_secs` 21 → 111 → 351+). Later `POST /turns` get HTTP 503 `retry_later`.

This is a **different** root cause from #759 (unbounded await / wall-clock abort). That fix cannot recover this path: the holder is stuck in a sync `parking_lot` lock, not an async wait.

### File:line cause (v0.2.26)

- `src-tauri/src/session_manager/connect.rs` ~416: `let mut guard = self.inner.lock();`
- ~433: `return Ok(self.snapshot());` while `guard` is still held
- `src-tauri/src/session_manager/process.rs` ~1226: `snapshot()` locks `self.inner` again → sync deadlock
- `connect.rs` ~87: `_connect_guard` never drops because `connect_inner` never returns

Same pattern at `connect.rs` ~348 (fork-busy early-return of `snapshot()` while holding `inner`).

`turn.rs` already documents the correct rule: while holding `inner`, return `snapshot_from_live` — never `self.snapshot()`.

## Steps to reproduce

1. Grok App **v0.2.26**, signed in with SuperGrok.
2. Cold Session API turn: `POST /v1/sessions/{id}/turns` — **succeeds** (spawn + first connect releases `inner` / `connect_lock`).
3. Second turn on the same Ready live session (already-live no-op).
4. Observe: second request returns **503** `retry_later`; `GET /v1/health` has `connectLockBusy: true`.

## Expected behavior

Already-live connect is a no-op that returns a snapshot and drops `connect_lock`. Second (and later) turns on a Ready live session succeed without wedging the host.

## App version

0.2.26

## Logs / screenshots

```
acp connect no-op: already live session=… state=Ready busy=false
```

No further connect return after that line.

`/v1/health`:

```json
{ "ok": true, "connectLockBusy": true }
```

Holder diagnostics: `connect_lock held by session=… phase=connect_inner since=21s` then `111s` then `351s+`.

First Session API turn (cold spawn) succeeds and releases the lock. Only the second turn on a Ready live session hits already-live and wedges.

## OS

Any (Host `parking_lot` deadlock; not OS-specific). Reproduced with Session API while signed in.

ynjmxn:
Fixes #905

## Problem
On the already-live path, `connect` held the `parking_lot` `inner` lock and then called `snapshot()`, which tries to lock `inner` again. `parking_lot` mutexes are not reentrant, so the thread self-deadlocked while still holding `connect_lock`. The first turn appeared to hang, and any second Session API turn deadlocked on `connect_lock` (surfacing as retry_later / no-op connect never returning).

## Fix
Take the snapshot from data already available under the held lock instead of re-locking: use `snapshot_from_live` / `snapshot_locked` in the already-live branch, so `inner` is locked exactly once and `connect_lock` is released normally.

The fork-busy branch had the same pattern (holding `inner` and calling `snapshot()`) and is fixed the same way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Fixed
+- **Session API second turn no longer deadlocks on already-live connect (#905)**: `connect_inner` held `inner` and then called `snapshot()`, which locks `inner` again (`parking_lot` is not reentrant). The thread never returned, so `connect_lock` stayed busy forever and later `POST /turns` got 503 `retry_later`. Already-live and fork-busy now snapshot from the held lock.
+
+**中文 · 修复**
+- **Session API 第二轮不再在 already-live connect 上死锁（#905）**：`connect_inner` 握着 `inner` 再调 `snapshot()`，而 `snapshot()` 会再锁一次（`parking_lot` 不可重入）。线程回不来，`connect_lock` 一直 busy，后续 `POST /turns` 变成 503 `retry_later`。already-live / fork-busy 改为从已持有的锁做 snapshot。
+
 ## [0.2.26] - 2026-08-24
 
 > **Highlight:** Import Grok Build CLI sessions from Account; phone mirror on the same Wi-Fi; Windows permission/rewind clicks; long chats stay scrollable.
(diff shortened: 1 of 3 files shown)


RongleCat:
Merged, thank you — this unblocks Session API second-turn connect. Follow-up: still worth a live already-live → second POST /turns check.

PR #936, opened 2026-08-27
fix(session): keep partial forks trimmed

a70win-wq:
## Summary

Fix partial conversation forks so later turns cannot be restored from the parent agent history while the child session is still being created.

## Root cause

`Fork from here` correctly truncated the App journal, but the new child temporarily retained the parent's agent session id. A deferred linked-history reconcile could run before child fork + rewind finished, merge the parent's full `chat_history`, and restore turns after the selected branch point.

The connect path also cleared the pending partial-fork metadata before the child rewind/bootstrap result and replacement agent id were persisted, widening that race.

## Changes

- Skip linked-agent history reconciliation while a partial fork is pending.
- Keep the pending trim metadata until the child agent id and rewind/bootstrap result are persisted together.
- On failed partial-fork connect, clear the inherited parent agent id as well as the one-shot flags, so a later reconnect bootstraps from the already-trimmed journal.
- Expand regression coverage to an 8-turn conversation forked through turn 5, plus pending/full reconcile and failed-connect lifecycle cases.

## Verification

- `pnpm typecheck`
- `pnpm dlx node@22 ./node_modules/vitest/vitest.mjs run` — 6,743 passed
- `pnpm build:ui`
- `pnpm deps:check`
- `rustfmt --edition 2021 --check` on the three changed Rust files
- Targeted Rust fork/reconcile/failure tests — 8 passed
- Full Rust library run — 1,522 passed, 1 ignored; one unrelated LAN-bind health test timed out because the local runner could not reach its own detected LAN address

Manual installed-app verification was not run; CI can exercise the repository matrix.

### DIFF
--- a/src-tauri/src/cli_sessions.rs
+++ b/src-tauri/src/cli_sessions.rs
@@ -1765,6 +1765,15 @@ pub fn reconcile_journal_from_chat_history(
     Ok(changed)
 }
 
+/// Linked reconcile may rebuild the App journal from agent `chat_history`.
+/// Skip while a **partial** fork is still pending (`fork_agent_session` plus a
+/// rewind index): the child journal is intentionally shorter than parent
+/// history, so a reconcile would overwrite the cut. Full forks and ordinary
+/// linked sessions stay eligible.
+fn linked_reconcile_allowed(meta: &SessionMeta) -> bool {
+    !(meta.fork_agent_session && meta.fork_rewind_prompt_index.is_some())
+}
+
 /// Best-effort reconcile when opening an App session that is linked to an agent.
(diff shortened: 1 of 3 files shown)


RongleCat:
已合入 [`f3acf3f4`](https://github.com/RongleCat/grok-app/commit/f3acf3f4)。谢谢 @a70win-wq。

审的时候补了一刀：rewind 失败且 `session/new` 也失败会提前返回，这条路径原先不会清 one-shot，下次连接仍可能带着父会话 agent id 重试 fork。见 [`a6801171`](https://github.com/RongleCat/grok-app/commit/a6801171)。

PR #948, opened 2026-08-28
fix(chat): settle bottom overscroll without hiding the tail

Johnny-dot:
## Summary

- preserve a recent wheel/touch intent toward the tail across WebView2 rubber-band rebound
- wait for a short quiet window, then recompute the latest `scrollHeight - clientHeight` and restore the bottom pin
- clear the rebound latch immediately on a real upward wheel/touch gesture so history browsing is never yanked back
- cover dynamic composer/queue height growth with pure-policy and DOM-hook regression tests

## Root cause

The floating composer changes the transcript's bottom padding through `ResizeObserver`. Near the tail, WebView2 can report elastic settle as a decreasing `scrollTop`. The scroll handler treated that reverse displacement as an intentional upward history gesture before the new composer padding had settled, set `escaped=true`, and stopped following. The viewport then rested above the true max while the last chat rows remained under the composer.

## Windows WebView2 runtime verification

Ran a side-by-side Tauri dev build with a distinct identifier, port, and data root. The runtime probe used the real `.lobe-chat__scroll`, `.lobe-chat__inner`, `.composer`, `ResizeObserver`, and wheel/scroll listeners:

- dynamic composer/queue-height probe changed max scroll position `1831 -> 1877`
- after downward overscroll + reverse settle: `scrollTop=1877`, bottom gap `0px`
- Back to latest button hidden after settle
- last content bottom `546px`, composer top `634px` (`88px` clear)
- explicit reverse upward wheel stayed at `max - 64px` and showed Back to latest

The isolated runtime process and WebView2 children were stopped after the smoke test.

## Automated verification

- `pnpm typecheck`
- `pnpm lint`
- `pnpm build:ui`
- focused scroll/virtualization suite: 128 passed, 0 failed
- changelog parser: 20 passed, 0 failed
- full frontend suite: 6,751 passed; the only three first-pass failures were two CRLF-sensitive source guards and one 5s parallel timeout
- LF guard rerun: 4 passed, 0 failed
- settings catalog isolated rerun: 16 passed, 0 failed

## Notes

The 320ms intent grace is longer than the 96ms quiet-settle window. A real direction reversal clears the grace synchronously in the wheel/touch handler, so the recovery path only applies to compositor rebound inside the existing near-bottom band.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,13 @@ See `docs/llm-wiki/release.md`.
 - Windows wallpaper frost reaches the sidebar and empty chat. WebView2 skipped container blur.
 - Fork from a middle turn no longer brings back later parent messages.
 - Rapid follow-up messages no longer lose the next reply.
+- Bottom overscroll no longer rebounds above hidden chat content.
 
 **中文 · 修复**
 - 修复 Windows 壁纸透过侧栏和空会话。WebView2 会跳过容器模糊。
 - 修复从中间分叉时后面的父会话消息又回来。
 - 修复连续快速发送时下一条回复丢失。
+- 修复底部超滚回弹后仍有聊天内容被遮挡。
 
 ## [0.2.27] - 2026-08-27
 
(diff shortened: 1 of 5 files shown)


RongleCat:
Landed on `main` via #949 (`046e957a`). Thank you @Johnny-dot — WebView2 bottom overscroll now settles without hiding the tail, and a real reverse gesture still escapes.

已合入 `main`。谢谢。

PR #951, opened 2026-08-28
fix(chat): follow latest tokens inside live thinking

### LINKED ISSUE #950: [Bug]: Long live thinking does not follow the latest tokens / 思考过长时看不到最新内容
### Describe the bug

When Grok App is thinking for a long time, the latest reasoning stays hidden. Short thoughts look fine. Once the chain of thought is taller than the thinking box, new tokens keep writing below the fold and you have to scroll the box by hand.

This is not the outer transcript stick-to-bottom bug (#931). #931 was about the chat list dropping its pin during stream. 0.2.27 fixed that path. Live thinking still lives in a nested scroller:

- `.grok-thought__body` is `max-height: min(32vh, 220px)` with `overflow-y: auto`
- `.grok-act__thought-body` is `max-height: 240px` with `overflow: auto`

The outer chat only sees a 220px box, so stick-to-bottom thinks it is already at the bottom. Nothing sets `scrollTop` on the inner box while tokens arrive. `Thinking.tsx` only opens and closes the fold.

**中文：** 思考一长，最新内容会停在思考盒子里看不见。短思考没事。这不是 0.2.27 修过的外层贴底（#931）。思考正文被盖在大约 220px 的内层滚动里，外层已经在底，内层没有跟滚。

Suggested labels: `bug`

### Steps to reproduce

1. Open Grok App 0.2.27 (or current `main`).
2. Stay at the bottom of a chat and send a prompt that produces a long live thought (high reasoning, many paragraphs of CoT).
3. Do not scroll.
4. Watch the thinking block once the text is taller than about 220px.

### Expected behavior

While thinking is live and the user has not scrolled up inside the box, the inner scroller should stay on the latest tokens. A user who scrolls up to read earlier reasoning should not be yanked back.

**中文：** 思考还在进行、用户没有在盒子里往上翻时，内层应跟到最新字。用户主动往上看时不要硬拽回来。

### App version

0.2.27 (also current `main`; this inner path was not changed after #931)

n/a (UI only)

### OS

Windows / macOS / Linux (CSS cap is not OS-specific)

### Logs / screenshots

Long live thinking cut off inside the capped box

```css
.grok-thought__body {
  max-height: min(32vh, 220px);
  overflow-y: auto;
}
.grok-act__thought-body {
  max-height: 240px;
  overflow: auto;
}
```

Do not use `Element.scrollIntoView` to fix this. That walks the chat ancestor and drops outer stick-to-bottom (documented next to `scrollChildIntoContainer`).

erict16:
## Summary

Fixes #950. Long live thinking grew inside a 220–240px inner scroller. Outer stick-to-bottom already sat on that box, so new CoT stayed below the inner fold unless you scrolled the box by hand.

## Changes

- Follow `scrollTop` on `.grok-thought__body` and `.grok-act__thought-body` while the thought is live and expanded.
- Leave the height cap in place. Do not use `scrollIntoView` (that walks the chat ancestor and drops outer pin).
- If the user scrolls up inside the box, stop following until they hit the inner bottom again.
- Contain overscroll on the bare thought body so it does not leak into the transcript.

## Verification

- `vitest run src/lib/thoughtBodyFollow.test.ts` — 8 passed
- CSS cap regression: `max-height: min(32vh, 220px)` / `240px` still present
- Both live surfaces wire `thoughtBodyRef` (`Thinking.tsx`, `TimelinePhaseBlock.tsx`)

I could not drive the installed Tauri window from here. The unit tests cover follow/escape math and the two call sites.

Screenshot: `docs/qa/live-thought-inner-scroll.jpg`


RongleCat:
Landed on `main` (`92eec70a`). Thank you @erict16 — live thinking now follows the latest tokens inside the capped box, without `scrollIntoView`.

已合入 `main`。谢谢。

PR #962, opened 2026-08-30
fix(rewind): bound agent RPCs and block rewind on background mid-turn chats

### LINKED ISSUE #961: [Bug]: Rewind agent RPCs are unbounded, and rewind ignores background mid-turn sessions
Three related defects in the rewind path on `main` (f5c4d38). All are host-side; found by reading the code against the fixes that already landed for #879 / #905, and each one has a sibling in the same file that already does the right thing.

## 1. `rewind_drop_last_user_turn` has no RPC budget

`src-tauri/src/session_manager/journal.rs` awaits `rewind_execute_for` with no timeout, and on a non-"method not found" error it awaits a **second** call at the fallback index:

- first attempt: `journal.rs:60`
- fallback attempt: `journal.rs:83`

`AcpClient::request` allows `HANDSHAKE_TIMEOUT_SECS` (45s) per method name and `rewind_execute_for` probes several names, so a slow or wedged agent can park the command for minutes with the rollback dialog spinning.

The sibling command in the same file already bounds the identical call at 8s:

- `journal.rs:232-236` — `tokio::time::timeout(Duration::from_secs(8), client.rewind_execute_for(...))`

**Impact:** matches the shape of #879 (rollback dialog hangs, cannot be dismissed).

## 2. Fork child rewind runs outside the handshake budget while holding `connect_lock`

`connect.rs:1250-1253` correctly wraps `initialize_and_open_session` in `with_handshake_budget`. The post-open child rewind is **not** wrapped:

- `connect.rs:1263-1266` — `client.rewind_execute_for(&agent_sid, prompt_index, false).await`

`connect_lock` is held for all of `connect_inner` (`connect.rs:90`), and other sends/connects queue on it (`turn.rs:60-62`, `session_api.rs:938-940`). So one fork-with-rewind can pin the global lock for the handshake budget **plus** the rewind probe chain.

**Impact:** other chats' send/connect stall; automations and the Session API return `RetryLater` / host-busy while one fork finishes. Note the repo already has `with_soft_rpc_budget`, documented precisely as "Soft-fail RPCs must not pin `connect_lock`" (`connect.rs:190-201`).

## 3. `rewind_to_prompt_index` busy check only reads the live slot

`journal.rs:192-207` computes `busy` only when `inner` holds a session whose `app_session_id` matches; every other case falls through to `(false, …, false)`. A chat demoted to `background` (`mod.rs:69-71`) while streaming therefore reports **idle**, and the local truncate at `journal.rs:271-273` runs anyway.

The check is also weaker than the authoritative helper: it tests only `fsm.state() == Streaming | AwaitingPermission`, while `live_session_is_busy` (`process.rs:268-275`) additionally honours `prompt_in_flight`, which stays true after an early `prompt_complete`.

`with_session_mut` (`process.rs:687-701`) exists for exactly this and its doc comment describes this bug class: *"that session may have been demoted when the user switched chats … so the background turn waited forever."*

**Impact:** switching away from a streaming chat and then rewinding it truncates the journal underneath a running turn, so the agent's memory no longer matches the transcript — the next send hangs or shows a phantom 「思考中」.

## Environment

- macOS (Apple Silicon), `main` @ f5c4d38
- Grok Build CLI 1.0.13

PR follows, with a regression test for (3) that fails on the live-slot-only read and passes with the fix.

ericyiu9819:
Fixes #961

Three defects in the rewind path. Each already had a sibling in the same file doing the right thing, so every fix reuses the existing in-repo idiom rather than inventing a new one.

## 1. Unbounded agent rewind RPCs (`journal.rs`)

`rewind_drop_last_user_turn` awaited `rewind_execute_for` with no timeout, and awaited a **second** unbounded call on the fallback index. `AcpClient::request` allows 45s per method name and `rewind_execute_for` probes several, so the rollback dialog could spin for minutes.

Both calls now use the 8s budget that `rewind_to_prompt_index` already applied inline, hoisted to `REWIND_AGENT_RPC_BUDGET` so all three call sites in the file share one number. A timeout on the first attempt **skips** the fallback index — spending a second budget would double a stall the user already felt.

## 2. Fork child rewind pinned `connect_lock` (`connect.rs`)

`initialize_and_open_session` is wrapped in `with_handshake_budget`, but the post-open child rewind was not, and `connect_lock` is held for all of `connect_inner`. One fork-with-rewind could therefore pin the global lock for the handshake **plus** the rewind probe chain, stalling every other chat's send/connect.

Now wrapped in `with_soft_rpc_budget`, whose doc comment states its purpose exactly: *"Soft-fail RPCs (set_model / set_mode) must not pin `connect_lock`."* A timeout lands on the existing `Err` arm, so the established `rewind_ok = Some(false)` fail-safe path handles it unchanged.

## 3. Rewind ignored background mid-turn chats (`journal.rs`)

`rewind_to_prompt_index` computed `busy` only when `inner` held a matching session; every other case fell through to `false`. A chat demoted to `background` while streaming reported **idle**, and the local truncate ran anyway — the agent's memory stopped matching the transcript, so the next send hangs or shows a phantom 「思考中」.

Extracted `rewind_blocked_by_running_turn`, which routes through `with_session_mut` (live **and** background — its doc comment describes this very bug class) and uses `live_session_is_busy`, which is authoritative where the FSM is not: it honours `prompt_in_flight` after an early `prompt_complete`.

`live_match` semantics are unchanged, so the agent-rewind path and the meta touch still apply only to the focused live chat.

## Tests

Two regression tests in `routing_tests_p2.rs`, using the existing `bare_live_session` fixture:

- `rewind_busy_check_sees_background_mid_turn_session` — seeds a demoted session with `prompt_in_flight = true` and asserts rewind is blocked. **Verified to fail** when `rewind_blocked_by_running_turn` is reverted to the live-slot-only read, and pass with the fix.
- `rewind_busy_check_allows_idle_background_session` — an idle demoted chat, and an unknown chat id, must not be blocked.

| Check | Result |
| --- | --- |
| `pnpm typecheck` | pass |
| `pnpm test` | 559 files, 6870 tests pass |
| `publish-website-downloads.py --self-test` | OK |
| `pnpm build:ui` | pass |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | 0 warnings |
| `cargo test` | 1582 pass, 0 fail (1580 before, +2 new) |

CHANGELOG updated under Unreleased → Fixed, en + zh, first sentences within the 90-character What's New limit (`whatsNew.test.ts` passes).

## Notes

- No behaviour change for the happy path: a responsive agent completes well inside 8s and takes the same branches as before.
- Only rewind paths are touched; no drive-by refactors.
- I did not change the 60s `with_soft_rpc_budget` value for the fork rewind — bounded by the existing idiom seemed better than introducing a new number. Happy to tighten it to a dedicated budget if you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ See `docs/llm-wiki/release.md`.
 - Chat no longer jumps while the agent is streaming. Switching away and back no longer lands in the middle.
 - Live thinking is not clipped in a 220px box. The transcript follows the stream.
 - Opening an SSH path chip no longer blanks the window. The files pane shows a loading state instead of an empty Suspense.
+- Rollback no longer spins for minutes when the agent is slow to answer. Forking a chat also stops holding up other chats' sends.
+- Rewind is refused while a chat you switched away from is still streaming. It used to cut that chat's history mid-turn.
 
 **中文 · 修复**
 - 壁纸上 CLI 更新条和「未打开标签」跟外观文字色。原先不在暴露层名单里，用了过淡的 tertiary。
@@ -40,6 +42,8 @@ See `docs/llm-wiki/release.md`.
 - 修复对话输出时屏幕往上跳、切出 App 再回来落到中间。正在看历史时不会被拽回底部。
 - 修复 Live Thinking 被 220px 小框截断。主对话跟着流式输出走。
 - 修复点 SSH 路径 chip 整窗变白。文件面板先显示加载，不再空白。
+- 修复回退在 agent 迟迟不应答时转好几分钟。分叉会话也不再占住别的会话的发送。
+- 修复切走的会话还在输出时仍允许回退。原先会截断那条会话的历史。
 
 ## [0.2.28] - 2026-08-29
 
--- a/src-tauri/src/session_manager/connect.rs
+++ b/src-tauri/src/session_manager/connect.rs
@@ -1261,9 +1261,16 @@ impl SessionManager {
 
                 match plan {
                     ChildTrimPlan::RewindChild { prompt_index } => {
-                        match client
-                            .rewind_execute_for(&agent_sid, prompt_index, false)
-                            .await
+                        // Post-open RPC, so it runs outside `with_handshake_budget`
+                        // while `connect_lock` is still held. Unbounded, it pinned
+                        // the lock for the handshake *plus* the rewind probe chain,
+                        // stalling every other chat's send/connect behind one fork.
+                        match Self::with_soft_rpc_budget(client.rewind_execute_for(
+                            &agent_sid,
+                            prompt_index,
+                            false,
+                        ))
+                        .await
                         {
                             Ok(_) => {
                                 rewind_ok = Some(true);
--- a/src-tauri/src/session_manager/journal.rs
+++ b/src-tauri/src/session_manager/journal.rs
@@ -11,7 +11,28 @@ use crate::store::{self};
 
 use super::*;
 
+/// Budget for a single agent rewind RPC. `AcpClient::request` already allows
+/// `HANDSHAKE_TIMEOUT_SECS` per method name and `rewind_execute_for` probes
+/// several names, so an unbounded await can park a rewind command for minutes
+/// with the rollback dialog spinning. The local journal is the UI source of
+/// truth, so exceeding the budget degrades to "agent rewind failed" instead.
+const REWIND_AGENT_RPC_BUDGET: std::time::Duration = std::time::Duration::from_secs(8);
+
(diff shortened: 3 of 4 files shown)


RongleCat:
Merged — thanks @ericyiu9819! The rewind RPC budget, `connect_lock` bounding, and background mid-turn block all reuse the existing in-repo idioms cleanly, and the regression tests fail on the live-slot-only read as described. Local CI is green (typecheck / 6870 tests / lint / code-quality gates / build:ui, plus cargo fmt + clippy + 1582 tests). 🎉

PR #964, opened 2026-08-30
fix(ask-user): restore a pending questionnaire after a missed one-shot emit

### LINKED ISSUE #963: [Bug]: Missed session://ask_user leaves the chat stuck thinking with no gate to answer
On `main` (f5c4d38) a pending agent questionnaire cannot be recovered, so losing the one-shot event strands the turn permanently. The permission card already has the symmetric fix; ask-user was never given one.

## Why it strands the turn

`session://ask_user` is emitted exactly once, from two places:

- live slot: `src-tauri/src/session_manager/events.rs:878-892`
- background slot: `src-tauri/src/session_manager/events_bg.rs:809-819`

Both store **only** the RPC id — `pending_ask_user_rpc_id` (`types.rs:197`). There is no stored payload, unlike permission which keeps the full `pending_permission_ui: Option<UiPermissionRequest>` (`types.rs:212`) precisely so it can be pulled again.

There is also no restore command. Permission has one:

- `SessionManager::pending_permission` (`control.rs:986-1003`)
- `session_pending_permission` (`commands/session_p1.rs:292-297`)
- pulled by the frontend on session open (`src/app/AppWorkbench.tsx:3507-3516`)

The ask-user branch immediately below only reads an in-window ref and never asks the host (`AppWorkbench.tsx:3517`).

Meanwhile the agent is blocked on the `_x.ai/ask_user_question` reverse RPC, and the stall watchdog **deliberately skips** sessions holding an ask gate (`watchdog.rs:91-92`), so nothing heals it either.

## Reproduce

1. Get the agent to ask a question (any turn that triggers `_x.ai/ask_user_question`).
2. Reload the WebView / remount the window while the questionnaire is on screen, or switch chats and back after a remount.
3. The questionnaire is gone, the chat shows 「思考中」 forever, and there is no control to answer. Only a force-quit or a new chat gets out.

This matches #844 ("The question popup for agent keeps freezing every time").

## Environment

- macOS (Apple Silicon), `main` @ f5c4d38
- Grok Build CLI 1.0.13

PR follows: stores the payload next to the rpc id, adds `session_pending_ask_user`, and has both gates share one restore helper so they cannot drift apart again.

ericyiu9819:
Fixes #963

`session://ask_user` is a one-shot emit while the agent blocks on the `_x.ai/ask_user_question` reverse RPC. Only `pending_ask_user_rpc_id` was stored and there was no restore command, so a WebView reload or window remount lost the questionnaire permanently — the chat sat at 「思考中」 with nothing to answer, and `watchdog.rs:91-92` deliberately skips sessions holding an ask gate, so nothing healed it either.

The permission card already had exactly this recovery path. Ask-user did not. This PR closes the asymmetry and then makes it structurally hard to reintroduce.

## Host

- `types.rs`: new `UiAskUserRequest` and `pending_ask_user_ui`, sitting next to `pending_ask_user_rpc_id` and documented the same way as `pending_permission_ui`.
- `events.rs` / `events_bg.rs`: populate it alongside the rpc id, then **emit the stored struct** instead of a separate `json!` literal — a live event and a restore can no longer drift apart in shape.
- `control.rs`: `pending_ask_user`, mirroring `pending_permission`, including the same invalidation rule — the read is gated on `pending_ask_user_rpc_id`, which is the field every clear path (`control.rs`, `turn.rs`, `process.rs`) already touches. So a resolved, recycled or stopped gate cannot resurrect, and no new clear sites were needed.
- `commands/session_p1.rs` + `lib.rs`: `session_pending_ask_user`.

`UiAskUserRequest` carries the real `Option<String>` / `Vec<AskUserQuestionItem>` types rather than `serde_json::Value`. With `rename_all = "camelCase"` the wire payload is byte-identical to the previous literal, so the existing `AskUserPayload` listener needs no change.

## Frontend

`restoreForSession` had the permission pull written inline and the ask-user branch reading only an in-window ref. Both now go through one `restoreSessionGate` helper in `src/lib/sessionGateRestore.ts`.

Behaviour is preserved exactly for permission: apply the parked gate synchronously, pull only when nothing is parked, ignore a reply for a chat the user already left, and never overwrite a live event that landed first (the helper re-checks the parked map before committing).

Per AGENTS.md item 7, the logic lives in a domain module and **App shell + AppWorkbench shrinks by 1 line** (15323 → 15322).

## Tests

- `pending_ask_user_recovers_questionnaire_until_invalidated` (Rust) — mirrors `pending_permission_recovers_card_until_invalidated`: recovers the questionnaire, refuses unknown session ids, and stops recovering once the rpc id is cleared.
- `sessionGateRestore.test.ts` — 6 cases: parked gate needs no host call; empty parked pulls and commits; a reply for a closed chat is dropped; a live event arriving mid-pull wins; disabled (non-Tauri) never calls the host; a rejected pull is swallowed.

| Check | Result |
| --- | --- |
| `pnpm typecheck` | pass |
| `pnpm test` | 560 files, 6876 tests pass (+6) |
| `publish-website-downloads.py --self-test` | OK |
| `pnpm build:ui` | pass |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | 0 warnings |
| `cargo test` | 1581 pass, 0 fail (+1) |

CHANGELOG updated under Unreleased → Fixed, en + zh, within the 90-character What's New limit.

## Notes

- No new clear sites: gating the read on the rpc id is the same contract `pending_permission` relies on, so every existing invalidation path already covers the new field.
- `session://ask_user_cleared` and the background attention signal are untouched.
- I kept `restoreSessionGate` generic rather than hardcoding the two payload types, so a future gate (plan mode) can reuse it.


RongleCat:
Thanks @ericyiu9819 — the fix is **merged into main** (9c1bdc0a, cherry-picked + rebased by a maintainer to resolve a CHANGELOG conflict with the rewind/rollback notes). The `pending_ask_user_ui` payload + shared `restoreSessionGate` helper are exactly the right shape; the read-gated-on-rpc-id invalidation mirrors `pending_permission` cleanly.

Verified locally on the merged tree: typecheck, 6876 frontend tests (incl. the 6 new `sessionGateRestore` cases), lint, code-quality gates, build, cargo fmt/clippy, 1583 Rust tests. Note: I trimmed the CHANGELOG bullet's first sentence to 85 chars to fit the 90-char What's New rule.

You can delete the `fix/pending-ask-user-restore` branch on your fork. 🎉

RongleCat:
Merged into main via 9c1bdc0a (maintainer cherry-pick + conflict resolution). Thanks for the fix!

Issue #853, opened 2026-08-23
[Perf]: Long chat transcript still stutters while scrolling after #842


zhangxaochen:
### Describe the bug

Follow-up to #842 / #843. Long transcripts still hitch while scrolling, especially when a large **Worked for …** activity block (many collapsed tool rows) enters the virtual window.

Typical symptoms:
1. **Lift-off freeze (~300ms+)** — finger/trackpad follow feels fine, then a pause before inertia, as if contact and fling are two separate gestures.
2. **Worked-for spike** — scrolling past a long activity phase drops frames (main-thread commits of 50–120ms) even though the rows are collapsed.
3. **Bottom bounce** — re-pinning / hitting the bottom makes the list jump while the pin window expands overscan.

#843 removed layout thrashing on row mount and `localStorage` hits. Remaining cost is mid-scroll **window expansion** (urgent React commits of many phase blocks at once), **hover style recalc** as content slides under a stationary pointer, **eager `toolExpandBody` (ANSI strip) on collapsed steps**, and **pin snap using post-commit metrics** that confuse height drift with the user leaving the bottom.

### Steps to reproduce

1. Open a long session with several **Worked for …** phases that each contain many tool steps.
2. Scroll the transcript with a precision touchpad (Windows WebView2, 120Hz).
3. Scroll across a large Worked-for block; lift off and let inertia run; flick back to the bottom.

### Expected behavior

Scroll stays on the gesture through lift-off and inertia. Collapsed Worked-for blocks do not take a 50ms+ commit when they enter overscan. Hitting the bottom does not bounce.

### App version

main after 0.2.25 / #843

### OS

Windows (WebView2). Likely any Chromium webview at 120Hz.

### Additional context

Profiled with DEV `scrollPerfDebug`: remaining severe frames were `phase_mount_tools*` batches on pin↔browse window swings, not glass overlays. Hover `:hover` recalc under a stationary cursor also showed up as idle-main-thread stutters.

PR #965, opened 2026-08-30
perf(chat): memoize the full-journal scans in the transcript render body

ericyiu9819:
Refs #853

Three full-journal scans sat in `ConversationThread`'s render body, so they ran on **every** render of the thread. The virtualizer re-renders the thread on each window shift, so a fling through a long chat paid all three O(n) passes per frame on top of the mount and measure work.

## Changes

- `hasStreamingAssistant` (`ConversationThread.tsx:2606`) moves into a `useMemo` on `[messages]`.
- The two scans inlined in the JSX at `:3159` — `isToolInlinedInAssistants(messages, …)` and a `messages.some(…)` looking for an existing tool row — collapse into one `liveToolHasRow` memo on `[messages, liveTool]`. `liveTool` is itself a `useMemo`, so its identity is stable while scrolling and the memo actually holds.

## Behaviour

Identical. The JSX predicate was `!(toolCallId && inlined) && !some(...)`, i.e. `!A && !B`; `liveToolHasRow` returns `A || B` and the JSX now reads `!liveToolHasRow`. The `toolCallId === undefined` case takes the same branches as before.

One small improvement falls out: the memo returns early when there is no live tool, so the scans are skipped entirely in the common case rather than running and finding nothing.

## Scope note

I looked at the `MessageNodeRail` scroll sync as well (`:157-170`) and deliberately left it alone. The `getBoundingClientRect()` calls there are consecutive reads with no interleaved writes, so they cost one layout flush per frame rather than one per row — much less than it first looks. Making it cheaper means either relying on `[data-message-id]` rows having monotonically increasing `top` in DOM order (which nested rows would break, and `pickActiveNodeIdFromRects` deliberately takes the *last* match), or plumbing the virtualizer's offset table into the rail. Both belong in their own change with a measurement to back them, not bundled here.

| Check | Result |
| --- | --- |
| `pnpm typecheck` | pass |
| `pnpm test` | 559 files, 6870 tests pass |
| `pnpm build:ui` | pass |

CHANGELOG updated under Unreleased → Changed, en + zh, within the 90-character What's New limit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,11 @@ See `docs/llm-wiki/release.md`.
 
 ### Changed
 - Missing translations are filled in. SSH, IM security, permission rules, and Doctor checks follow the UI language.
+- Scrolling a long chat does less work per frame. Three full-journal scans no longer run on every render.
 
 **中文 · 变更**
 - 补上缺的翻译。SSH、IM 安全、权限规则和 Doctor 检查跟界面语言。
+- 长会话滚动每帧少做一些活。三处全量遍历不再每次渲染都跑。
 
 ### Fixed
 - CLI update banner and empty side-tab hint follow Appearance text color on wallpaper. They sat outside the old chrome list and used tertiary ink.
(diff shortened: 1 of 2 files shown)


RongleCat:
Thanks @ericyiu9819 — merged! The memoized scans are a clear win for long-chat scrolling. Appreciate the note on why the rail scroll sync was left alone (measured, not bundled). 🎉

PR #967, opened 2026-08-30
perf(chat): patch row offsets in place instead of rebuilding the prefix sum

ericyiu9819:
## Problem

`commitRowHeight` already knows which row changed and by how much, but it throws away the entire offsets cache:

```ts
heightsVersionRef.current += 1;
offsetsCacheRef.current = null;
```

The next `getOffsets()` therefore re-derives the whole table through `cumulativeOffsets(count, getHeight)`, and `getHeight` is not cheap per row — a `Map` lookup, then an `estimateRef` callback, then the fallback chain.

The commits that matter arrive in bursts: a `ResizeObserver` batch when a window of markdown rows mounts. Each commit in the burst triggers one full rebuild of a table whose length is the whole transcript.

## Fix

`offsets[i]` is the sum of the heights below `i`. Changing one row leaves every offset up to and including that row untouched and moves the whole suffix by the same amount, so the cache can be patched instead of rebuilt — no per-row height lookup, and only the rows after the edited one are touched. Commits near the end of a long thread (the common case while reading recent messages) touch almost nothing.

### The rounding trap

The shift is deliberately **not** `nextH - prevH`. `getHeight` returns the estimate *unrounded* for a row that has never been measured:

```ts
if (est != null && Number.isFinite(est) && est >= 0) return est;
```

while `commitRowHeight` rounds it into `prevH`. With a 120.4px estimate the naive delta is off by 0.4px, and since each patch feeds the next, the error accumulates down the suffix. The shift is instead derived from the height already recorded in the cached table, which is exact by construction regardless of rounding.

Falls back to the old invalidation when `index` is out of the cached table's range.

## Behavior

None intended. The patched table is value-identical to a rebuild; only the cost of producing it changes. `shiftOffsetsAfter` copies rather than mutating, so callers holding the previous array are unaffected.

## Tests

4 cases on `shiftOffsetsAfter` in `src/lib/chatVirtualList.test.ts`, including equivalence with a full `cumulativeOffsets` rebuild and the accumulating-drift case that pins the derivation above.

## Local CI

- `pnpm typecheck` clean
- `pnpm test` — 6874 passed / 559 files
- `pnpm lint` clean (`--max-warnings 0`)
- `whatsNew` CHANGELOG gate green

No Rust files touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,11 @@ See `docs/llm-wiki/release.md`.
 
 ### Changed
 - Missing translations are filled in. SSH, IM security, permission rules, and Doctor checks follow the UI language.
+- Long chats scroll more smoothly. Measuring one row no longer re-adds every row above it.
 
 **中文 · 变更**
 - 补上缺的翻译。SSH、IM 安全、权限规则和 Doctor 检查跟界面语言。
+- 长会话滚动更顺。测量一行不再把它上面所有行的高度重新加一遍。
 
 ### Fixed
 - CLI update banner and empty side-tab hint follow Appearance text color on wallpaper. They sat outside the old chrome list and used tertiary ink.
--- a/src/hooks/useChatMessageVirtualizer.ts
+++ b/src/hooks/useChatMessageVirtualizer.ts
@@ -40,6 +40,7 @@ import {
   chatOpenPinWindow,
   computeChatVirtualWindow,
   cumulativeOffsets,
+  shiftOffsetsAfter,
   resolveChatOverscanPx,
   shouldCommitRowHeight,
   shouldVirtualizeChat,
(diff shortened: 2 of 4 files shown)


RongleCat:
Thanks @ericyiu9819 for this clean perf fix! 🙌

The patch-in-place approach is correct — deriving the shift from the height already recorded in the offsets table (instead of `nextH - prevH`) is exactly what avoids the estimate-rounding drift, and the equivalence + fractional-drift tests pin that down nicely.

Merged onto `main` as `64b2d0ae`. I resolved a CHANGELOG conflict with the other in-flight scroll-perf bullet from #965 (both kept). Local CI green: `pnpm typecheck`, 6880 tests, `pnpm lint`, and the code-quality gates.

Thanks again for the contribution!

Issue #697, opened 2026-08-19
[Bug] 工具轮结束后正文不出现，切走再切回才显示 / Assistant body missing after tool turn until remount


erict16:
### 现象 / What happens

工具很多的一轮跑完后，助手正文有时不出现。侧栏已经 Ready，工具行也在，最后那段话没有。切到别的会话再切回来，journal 里的正文才画出来。

字已经写进磁盘了。当前正在看的这条会话，没把收尾 token / journal 尾巴刷进 UI。

After a long tool turn, the assistant body sometimes never paints. Sidebar is Ready and the tool rows are there, but the final answer is missing. Switch to another session and back: the journal body appears.

The text is on disk. The live view never applied the tail.

### 复现 / Repro

单测（不用开 App） / Unit test (no App needed):

`src/lib/streamLateToken.test.ts` — `applies late answer after a tool-only bubble was settled by early ready`

路径 / Path:

1. 工具轮把助手气泡结算掉（`streaming=false`，`content=""`，没有 thought）。 / A tool turn settles the assistant bubble (`streaming=false`, `content=""`, no thought).
2. `session://stream` 的 `done` 把 liveMap 翻成 `ready`。 / `session://stream` `done` flips liveMap to `ready`.
3. 真正的回答 token 在这之后才到。 / The real answer tokens arrive after that.
4. 旧逻辑：`shouldApplyLateStreamText` 把工具轮后的空正文当成已结束，把 chunk 丢掉。 / Old logic: `shouldApplyLateStreamText` treated a tool-only empty body as finished and dropped the chunk.
5. 后面的 `session://state` `ready` 是 ready→ready，`isTurnDoneReadyTransition` 为 false，journal heal 不跑。 / Later `session://state` `ready` is ready→ready, so `isTurnDoneReadyTransition` is false and journal heal does not run.
6. Host 如果已经写过这一行，post-turn reconcile 报 `changed == 0`，也不会发 `session://journal_reconciled`。 / If Host already wrote the row, post-turn reconcile reports `changed == 0` and never emits `session://journal_reconciled`.
7. 切走再切回会跑 `openSession` 从磁盘读，正文出现。 / Switch away and back runs `openSession` from disk. The body appears.

[...]

PR #969, opened 2026-08-31
fix(session): paint the answer instead of leftover thinking

### LINKED ISSUE #968: [Bug]: Thinking stays on screen as the final reply until you leave the session
Suggested labels: `bug` / `priority:p1` / `area:session`

### Describe the bug

Sometimes when a turn finishes thinking, the real answer is already in the session, but the chat still shows the thinking text as the final reply. Leave the session and open it again and the answer is there, folded correctly.

This is the same family as #697 (late body missing until remount). That fix covered empty thinking-only / tool-only bubbles. This path is different: live `segments` stay thought-only (or thought leaked into `content`) while the answer already sits on the message field / journal. Paint uses segments, so CoT looks like the reply until remount rebuilds from disk.

**中文：** 思考结束了，最终回复其实已经有了，界面却把刚才的思考过程当成最终回复。退出会话再点进来就正常。#697 修过「空泡不画正文」，这条是直播时间线还停在思考上。

### Steps to reproduce

1. Open Grok App 0.2.29 (or current `main`).
2. Send a prompt that thinks for a while, then writes a real answer (high reasoning, optional tools).
3. Wait until thinking ends. Do not switch sessions.
4. If the bubble still shows the reasoning as the reply, switch to another session and back.

Intermittent. Easier on long thinking turns where the body lands around the same time Host goes `ready`.

### Expected behavior

When thinking ends, the answer should paint in that bubble. Thinking collapses to “Thought for Ns”. Switching sessions should not be required.

**中文：** 思考一结束，正文应出现在同一条气泡里。思考折成「思考了 Ns」。不用切走再切回来。

### App version

0.2.29 (also current `main`)

n/a (UI projection)

### OS

Windows / macOS / Linux (not OS-specific)

erict16:
## Summary

Fixes #968. After thinking ends, the real answer can already be on the message / journal while the live timeline still paints CoT as the reply. Switching sessions remounts from disk and looks right. Same family as #697; this path is thought-only live `segments` (or thought leaked into `content`) winning over the body field.

Paint uses `messageSegments`. If `segments` is thought-only, the content field is ignored. Weave can also derive-wipe the body. Late assistant tokens after settle used to mint a second bubble or get dropped. Stream `done` could flush before pending thought.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## What changed

- `messageSegments` / `ensureSegments` append a content segment when the field has a real answer and live segs do not.
- Late body tokens bind to the settled current-turn assistant. Late thought does not re-open 思考中.
- Journal heal still lifts when live content *is* the thought, even if it is longer than the journal answer. Heal window covers Host post-turn reconcile.
- Stream coalescer drains the other kind (thought vs assistant) before a `done` tick.

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (targeted session/stream tests; 133 passed)
- [ ] I ran `cargo test` in `src-tauri` (no Host change)
- [x] User-facing strings go through `src/i18n/messages.ts` (en + zh)
- [x] Docs / `docs/llm-wiki` updated if behavior changed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ See `docs/llm-wiki/release.md`.
 
 **What's New popup:** each bullet's **first sentence** is what users see. Keep that sentence to one short line (added / fixed / improved — no paths, no implementation, no issue piles). A second short sentence is allowed for GitHub / this file only. Do **not** rewrite already-shipped `## [X.Y.Z]` sections.
 
+## [Unreleased]
+
+### Fixed
+- Thinking no longer stays on screen as the final reply. The real answer paints in place; switching chats is not required (#968).
+
+**中文 · 修复**
+- 思考结束后不再把思考过程当成最终回复。正文会直接画出来，不用切走再切回来（#968）。
+
 ## [0.2.29] - 2026-08-31
 
 > **Highlight:** One-click CLI install, complete translations, and smoother long chats.
--- a/docs/qa/issue-thinking-as-final-reply.md
+++ b/docs/qa/issue-thinking-as-final-reply.md
@@ -1,0 +1,36 @@
+Suggested labels: `bug` / `priority:p1` / `area:session`
+
+### Describe the bug
+
+Sometimes when a turn finishes thinking, the real answer is already in the session, but the chat still shows the thinking text as the final reply. Leave the session and open it again and the answer is there, folded correctly.
+
+This is the same family as #697 (late body missing until remount). That fix covered empty thinking-only / tool-only bubbles. This path is different: live `segments` stay thought-only (or thought leaked into `content`) while the answer already sits on the message field / journal. Paint uses segments, so CoT looks like the reply until remount rebuilds from disk.
+
+**中文：** 思考结束了，最终回复其实已经有了，界面却把刚才的思考过程当成最终回复。退出会话再点进来就正常。#697 修过「空泡不画正文」，这条是直播时间线还停在思考上。
+
+### Steps to reproduce
+
+1. Open Grok App 0.2.29 (or current `main`).
+2. Send a prompt that thinks for a while, then writes a real answer (high reasoning, optional tools).
+3. Wait until thinking ends. Do not switch sessions.
+4. If the bubble still shows the reasoning as the reply, switch to another session and back.
+
+Intermittent. Easier on long thinking turns where the body lands around the same time Host goes `ready`.
+
+### Expected behavior
+
+When thinking ends, the answer should paint in that bubble. Thinking collapses to “Thought for Ns”. Switching sessions should not be required.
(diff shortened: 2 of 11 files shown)


shiaho777 (commented):
Reviewed the diff end to end — this is a clean, well-tested fix and I'd like to see it merged.

**Verified locally (main `d4f686be`):**
- `pnpm typecheck` ✅ · `pnpm lint` ✅ · `pnpm test` 560 files / 6880 tests ✅
- Applied the PR's source changes on top of main in a scratch worktree: full `pnpm test` still passes (all 6880 + the new `session.thinkingAsReply.test.ts` cases).

**What I like:**
- Correct root-cause layering: the `contentLooksLikeThought` / `syncContentIntoSegments` guard in `messageSegments` / `ensureSegments` fixes the paint path once, rather than patching each caller, and it correctly refuses to promote leaked CoT into a content segment (nice test).
- The `StreamCoalescer` cross-kind drain on `done` and the `keepThoughtStreaming` guard both close real re-open/duplicate-bubble paths; tests cover both.
- The `JOURNAL_REHYDRATE_RETRY_GAPS_MS` extension to `[400, 500, 800]` matches Host's 750ms POST_TURN_RECONCILE — the comment documents the invariant, good.
- The two `docs/qa` markdown files are new (the folder previously held only screenshots); they read like the issue/PR bodies duplicated. Minor: consider dropping them from the PR or trimming to a short repro-notes file, to keep `docs/qa` for artifacts.

One behavior question, non-blocking: `findCurrentTurnAssistantForStream` binds to *any* non-error assistant row above the last user row, even one from a settled earlier interjection within the same turn. The bind-floor guard in the `idx > bindFloor` branch partially covers it, but if a turn has assistant → user interjection → late thought, the late thought will bind to the interjection's row. Harmless in practice (interjections sit after the main bubble), just flagging it in case a follow-up tightens it to the *last* assistant row.

Nice work — awaiting maintainer merge.

shiaho777:
I verified your changes against current main — details in my review above. Typecheck, lint, and the full test suite pass with your patch applied. The core paint-path fix (segments + coalescer drain) looks correct to me. Left one non-blocking question about late-thought binding to interjection rows. Maintainers will make the merge call.

RongleCat:
感谢 @erict16！PR #969 已合并。根因抓得很准：live `segments` 停在 thought-only（或 CoT 泄漏进 `content`），而正文其实已在字段/journal 上——分层修复 + 回归测试干净地关掉了 #968。社区评审与 CI 均绿。🎉

Thanks @erict16 — **#969 is merged**. Sharp root-cause work on leftover thinking paint; closes #968. Feel free to delete the branch on your fork.

PR #981, opened 2026-08-31
fix(ui): replace the last two native <select> dropdowns with the project Select

shiaho777:
## Summary

`docs/llm-wiki/dialogs.md` forbids OS-default controls for product actions; the codebase had already migrated everywhere except two stragglers still rendering browser-native `<select>` dropdowns inside modals:

- `DoctorModal` — the findings category filter
- `AgentDashboardModal` — the dispatch target project picker

## What changed

Both now render the project's portaled [`Select`](src/components/Select.tsx) component (`menu-panel` + `listbox`, menu portaled to body so it is never clipped by modal overflow, same chrome as every other settings dropdown).

- Doctor filter keeps the per-category count suffixes `(n)` and its `aria-label`; the `<label>`+`sr-only` wrapper collapses into the Select's own accessible labeling.
- Dashboard picker keeps the dispatch-hint reset on change and its `aria-label`; the `htmlFor`/`id` pairing becomes a labeled span + Select.

## Test plan

- [x] `pnpm typecheck` ✅ · `pnpm lint` ✅
- [x] `npx vitest run src/components` — 209 passed
- [x] Repo-wide grep: zero native `<select>` outside comments remain
- [x] No copy changes — all strings still flow through i18n

## Type of change
- [x] Bug fix (dialog-policy compliance)
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/src/components/AgentDashboardModal.tsx
+++ b/src/components/AgentDashboardModal.tsx
@@ -7,6 +7,7 @@ import { useEffect, useMemo, useState } from "react";
 import type { Locale, MessageKey } from "@/i18n";
 import { createT } from "@/i18n";
 import { GlassModal } from "@/components/GlassModal";
+import { Select } from "@/components/Select";
 import { xEvidenceStats, type XEvidenceStats } from "@/lib/api";
 import {
   AGENT_DASHBOARD_STATUS_FILTERS,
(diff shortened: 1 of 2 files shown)


RongleCat:
谢谢 @shiaho777！把 Doctor 分类筛选和 Agent Dashboard 派发项目选择换成项目 `Select`，清掉最后两处原生 `<select>`，符合 dialogs 规范。CI 全绿，已合并 🙏

PR #983, opened 2026-08-31
fix(ui): stop expanded tool steps stacking title and command

Ksndj:
## Summary
Expanded tool rows in the Worked-for list stacked title, command, and child steps on one line (garbled last row on Windows WebView2).

Root cause: `.grok-act__step` uses `content-visibility: auto` with `contain-intrinsic-size: auto 36px`. After expand, WebView2 kept the collapsed 36px box and painted the body over the label.

## Changes
- Expanded / speech steps force `content-visibility: visible` and real height (including the virtual 36px override).
- Bare `TimelineToolRow` gets `is-expanded` when the body is open.
- Guard test + CHANGELOG Unreleased.

## Test plan
- [x] `pnpm typecheck`
- [x] Related vitest (`chatAnswerOverlap.guard`, `whatsNew`, `grokActivityVirtualize`)
- [x] `pnpm lint`
- [x] `python scripts/check-code-quality-gates.py --mode final`
- [x] `python scripts/publish-website-downloads.py --self-test`
- [x] `pnpm build:ui`
- [x] `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` in `src-tauri`
- [ ] Full `pnpm test`: 2 pre-existing Windows failures (CRLF source-guard matches in `welcomeIntro.guard` / `extensionsSurface.guard`); not from this change. Linux CI is the authority.
- [ ] `cargo test`: Windows `STATUS_ENTRYPOINT_NOT_FOUND` (pre-existing); clippy/fmt clean. Linux/mac CI runs tests.
- [ ] Manual: expand the last tool in a Worked-for list — title on one line, command/output below, no stacked glyphs.

## 中文
展开的工具步骤不再把标题和命令叠在同一行。工作列表最后一行会按真实高度排开。

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,14 @@ See `docs/llm-wiki/release.md`.
 ## [Unreleased]
 
 ### Fixed
+- Expanded tool steps no longer stack title and command on one line. The last row in a Worked-for list keeps its real height.
 - Thinking no longer stays on screen as the final reply. The real answer paints in place; switching chats is not required (#968).
 - Settings search finds Chinese keywords for permission, telemetry, login, and more. Bilingual keywords stay locked by catalog test (#970).
 - Queue edit and other glass dialogs stay above the embedded browser. Native webviews hide while the modal is open (#976).
 - Doctor and the agent dashboard no longer use native dropdowns. They use the same app Select as settings (#981).
 
 **中文 · 修复**
+- 展开的工具步骤不再把标题和命令叠在同一行。工作列表最后一行会按真实高度排开。
 - 思考结束后不再把思考过程当成最终回复。正文会直接画出来，不用切走再切回来（#968）。
 - 设置搜索补上了中文关键词，权限、遥测、登录等能搜到了。目录测试会锁住双语关键词（#970）。
 - 打开内置浏览器时，队列编辑等毛玻璃弹窗不再被挡住。弹窗打开期间会暂时藏起原生页面（#976）。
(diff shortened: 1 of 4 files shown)


shiaho777 (commented):
Reviewed the diff — root cause and fix both look right.

**Verified locally (main `a024d1fb`):**
- `pnpm typecheck` ✅ · `pnpm lint` ✅
- `npx vitest run src/lib/chatAnswerOverlap.guard.test.ts src/lib/whatsNew.test.ts src/lib/grokActivityVirtualize.test.ts` (with the PR's patch applied in a scratch worktree) — all pass, including the three new guard assertions.

**Assessment:**
- The `content-visibility: auto` + `contain-intrinsic-size: auto 36px` collision on expanded rows is a genuine WebView2 paint hazard, and forcing `content-visibility: visible` + real height on `.is-expanded` / `[data-expanded="1"]` / `--speech` covers all three paths that open a body (including the virtualized container override).
- The guard test pins all three selectors so the intrinsic-size override can't silently regress.
- CHANGELOG bullet follows the What's New first-sentence rule.

One tiny nit, non-blocking: `TimelineToolRow` now sets both the `is-expanded` class and the `data-expanded="1"` attribute from the same `expanded` state, so the CSS selector list has a redundant arm (the attribute selector matters only for rows rendered by other code paths that set the attribute without the class — worth keeping if that's the intent, e.g. speech rows elsewhere; just flagging in case a future cleanup wants to collapse them).

Good catch on the Windows-only garbling — awaiting maintainer merge.

RongleCat:
谢谢 @Ksndj！首次贡献就踩中 WebView2 `content-visibility` 展开后仍锁 36px 的坑，CSS + `is-expanded` + guard test 很清楚。已批准并跑通 CI，squash 合并。欢迎继续提 PR 🙏

PR #988, opened 2026-09-01
perf(workbench): pause the git dirty poll while the window is hidden

shiaho777:
## What

The composer dirty chip polls `git_status` every 2–8s for the active project. The interval kept firing while the window was hidden (minimized / covered / another Space), spending a git process spawn + porcelain parse per tick on a chip nobody can see.

## How

- New shared helper `src/lib/visibilityPoll.ts`: interval ticks no-op while `document.visibilityState === "hidden"`, and one catch-up refresh fires on becoming visible (so data changed while hidden is not a full period stale).
- The dirty-poll effect in `AppWorkbench` now uses it; the focus listener is unchanged. Net effect: zero `git status` spawns while hidden, same freshness while visible.

## Checks

- [x] `pnpm typecheck`
- [x] `pnpm test` (6893 passed)
- [x] `pnpm lint`
- [x] `python3 scripts/check-code-quality-gates.py --mode final`
- [x] `python3 scripts/publish-website-downloads.py --self-test`
- [x] `pnpm build:ui`
- [x] New unit tests in `src/lib/visibilityPoll.test.ts`

No user-facing strings; no UI chrome changes.

### DIFF
--- a/src/app/AppWorkbench.tsx
+++ b/src/app/AppWorkbench.tsx
@@ -606,6 +606,7 @@ import {
   summarizeGitDirty,
   type GitDirtySummary,
 } from "@/lib/workspaceGit";
+import { startVisibilityPoll } from "@/lib/visibilityPoll";
 
 const AutomationsPage = lazy(async () => {
   const m = await import("@/components/AutomationsPage");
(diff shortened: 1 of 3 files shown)


RongleCat:
谢谢 @shiaho777！窗口隐藏时停掉 git dirty 轮询，露出来再补一票，省进程又保新鲜度。CI 全绿，已 squash 合并 🎉

PR #989, opened 2026-09-01
perf(chat): make the transcript open-reveal settle event-driven

shiaho777:
## What

When opening a historical session with media, the transcript is kept hidden until image/video cards have a measurable box, then scrolls once and reveals. While holding, a `setInterval(consider, 80)` swept the DOM 12.5×/s for up to 8s — per session open — competing with the very first paint this path is meant to protect.

## How

The sweep already has event drivers: frames flip `is-pending` → `is-ready`/`is-broken` classes (captured by the existing `MutationObserver` on `class`/`src`), `` fires `load`/`error`, `<video>` fires `loadedmetadata`, and cards mount/unmount (childList). The interval is only a safety net for anything that slips through, so it now idles at **500ms** instead of 80ms. New constant `TRANSCRIPT_OPEN_REVEAL_FALLBACK_POLL_MS` with a test locking it to ≤ timeout/10.

Reveal latency is unchanged in the common case (events fire `consider()` immediately); worst case for an event-less straggler settles within 500ms, still far below the 8s timeout.

## Checks

- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm lint`
- [x] `python3 scripts/check-code-quality-gates.py --mode final`
- [x] `pnpm build:ui`

No user-facing strings; no behavior change to when the reveal finishes.

### DIFF
--- a/src/components/lobe-chat/ConversationThread.tsx
+++ b/src/components/lobe-chat/ConversationThread.tsx
@@ -110,6 +110,7 @@ import {
   shouldHoldTranscriptOpenReveal,
   transcriptOpenRevealHasMedia,
   transcriptOpenRevealSettleMs,
+  TRANSCRIPT_OPEN_REVEAL_FALLBACK_POLL_MS,
   TRANSCRIPT_OPEN_REVEAL_TIMEOUT_MS,
 } from "@/lib/transcriptOpenReveal";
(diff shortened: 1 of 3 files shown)


RongleCat:
谢谢 @shiaho777！打开历史会话的 media reveal 改成事件驱动 + 500ms 兜底，首屏少扫很多 DOM。已合并 🙏

PR #991, opened 2026-09-01
perf(host): move the remaining git IPC family onto the blocking pool

shiaho777:
Follow-up to #990. Extends the same spawn_blocking pattern to the rest of the git IPC surface in `commands/git_p1.rs`.

## What

All of these are `async fn` Tauri commands with process spawns / disk writes and no intervening await — they ran directly on the async runtime worker, so a large workspace blocked every other queued command:

- `git_file_diff` — git probe + diff + untracked `--no-index` fallback (Changes panel per-file diff)
- `git_review_bundle` — 5 git spawns + numstat/patch parsing (Review tab bulk load)
- `git_show_file` — `git show HEAD:path` + binary sniff
- `git_checkout_file` — `ls-files` probe, untracked delete, `checkout`/`restore` (diff reject/restore)
- `apply_file_patch` — up to 2MB temp-write + rename (Changes Accept / Restore)

## How

Fast-path guards (path normalization, existence, project-root checks) stay on the runtime; each blocking body becomes a `*_blocking(project, …)` fn wrapped with `tauri::async_runtime::spawn_blocking`, same as `git_status` (#990) and `cli_worktrees_list`. Responses and soft-fail reasons are unchanged.

## Checks

### DIFF
--- a/src-tauri/src/commands/git_p1.rs
+++ b/src-tauri/src/commands/git_p1.rs
@@ -16,6 +16,8 @@ pub async fn git_file_diff(
     project_path: String,
     path: String,
 ) -> Result<GitFileDiffResult, String> {
+    // Fast guards on the runtime; the diff process spawns run on the blocking
+    // pool so a big diff cannot stall other queued commands.
     let project = normalize_fs_path(&project_path);
     let target = normalize_fs_path(&path);
     if project.is_empty() || target.is_empty() {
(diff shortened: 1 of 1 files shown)


RongleCat:
Merged — thanks @shiaho777! Extending the spawn_blocking pattern across the rest of the git IPC family keeps large diffs / review loads from stalling the Host runtime.

PR #992, opened 2026-09-01
perf(host): move worktree IPC commands onto the blocking pool

shiaho777:
Follow-up to #990 / #991, same `spawn_blocking` pattern for `worktree_agents_p1/p2.rs`.

## What

- `git_worktree_add` — porcelain list, `mkdir -p`, `git worktree add` (materializes a full checkout), re-list
- `git_worktree_gc` — repo-wide prune scan + prune
- `git_worktree_remove` — deletes an entire checkout tree
- `git_worktree_compare` — cross-worktree `git diff --name-status`

All four ran on the async runtime worker, so creating/removing a worktree from the UI froze every other queued IPC command for the duration of the disk walk.

## How

Fast-path guards (path checks, name/ref sanitize, probe) stay on the runtime; spawns and disk walks move to `*_blocking` fns behind `tauri::async_runtime::spawn_blocking`. `git_push_branch` and `gh_pr_create` already used `spawn_blocking` and are untouched.

## Checks

### DIFF
--- a/src-tauri/src/commands/worktree_agents_p1.rs
+++ b/src-tauri/src/commands/worktree_agents_p1.rs
@@ -647,6 +647,21 @@ pub async fn git_worktree_add(
     let start = sanitize_worktree_ref(start_point.as_deref())?;
     let layout_kind = normalize_worktree_layout(layout.as_deref());
 
+    // Worktree add materializes a full checkout (spawn + disk-heavy) — the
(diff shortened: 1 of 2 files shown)


RongleCat:
Merged — thanks again @shiaho777! Worktree add/remove/gc/compare on the blocking pool is the right follow-up so UI worktree ops cannot freeze other IPC.

PR #993, opened 2026-09-01
perf(host): run git_worktrees_list on the blocking pool

shiaho777:
Follow-up to #990 / #991 / #992 — closes out the git IPC family.

## What

The composer branch chip calls `git_worktrees_list` whenever the git menu can appear. `git worktree list --porcelain` stats every registered worktree on disk. That ran directly on the async runtime worker, so on a slow disk / large repo the list delayed every other queued command.

## How

`git worktree list` + porcelain parsing move to `git_worktrees_list_blocking` behind `tauri::async_runtime::spawn_blocking`; fast guards (empty path, not a directory, probe) stay on the runtime. Response shape unchanged.

## Checks

### DIFF
--- a/src-tauri/src/commands/git_p2.rs
+++ b/src-tauri/src/commands/git_p2.rs
@@ -291,6 +291,16 @@ pub async fn git_worktrees_list(project_path: String) -> Result<GitWorktreesResu
         });
     }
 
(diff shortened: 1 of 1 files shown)


RongleCat:
Merged — thanks @shiaho777! Composer branch-chip worktree list on the blocking pool closes out this git IPC series cleanly.

Issue #1003, opened 2026-09-02
[Bug]: macOS 0.2.19 can stay blank after in-app CLI upgrade until the view is recreated


toby-bridges:
### Describe the bug

On the signed macOS ARM64 build of Grok App `0.2.19`, the app entered a persistent blank window after its in-app Grok Build CLI updater upgraded the CLI from `1.0.5` to `1.0.13` and recycled the agent process.

When inspected, the WebView was still navigated to:

```text
tauri://localhost#/settings/remote_im/im
```

The native app, WebContent process, and CLI/ACP connection were still alive, but the window contained no rendered controls and the macOS accessibility tree had no child UI nodes below the WebView. `Cmd+R` did not recover it.

Leaving Settings with `Cmd+,` recreated the main workbench, and reopening Settings rendered the same route normally. A full quit/relaunch also rendered normally with the same App `0.2.19` + CLI `1.0.13` combination.

This was one observed incident, not yet a deterministic clean-room reproduction. The CLI update is temporally associated with the blank state, but the evidence does **not** prove that the CLI update alone caused it. The desktop App bundle itself was not updated.

This looks related to the Settings rendering work in #848, which shipped in `v0.2.25`, but it was more than the brief blank frame described there: the installed `v0.2.19` window stayed blank until the view was recreated. It may also expose a recovery/compatibility gap because an older signed App can independently self-update to the newest CLI.

### To reproduce

This is the observed sequence; it is not yet known which state is required to reproduce it consistently.

1. Run the signed Grok App `0.2.19` on Apple Silicon for an extended period.
2. Use the in-app CLI updater. It runs `~/.grok/bin/grok update` and upgrades Grok Build CLI from `1.0.5` to `1.0.13`.
3. The app recycles its agent process with `reason=cli_upgrade`.
4. Navigate or return to Remote Control > IM (`#/settings/remote_im/im`).
5. Observe that the entire app window is blank. `Cmd+R` leaves it blank.
6. Press `Cmd+,` to leave Settings, or fully quit/relaunch the app. Rendering returns; the same Settings route can then render normally.

### Expected behavior

- The workbench and Settings UI should continue rendering after a CLI upgrade and agent recycle.
- If the frontend root fails to render, the app should show an error boundary with a retry/reload action instead of a permanent blank window.
- If a newer CLI requires a newer App, the in-app CLI updater should warn, gate the update, or direct the user to update the App first.

### Version

[...]

PR #1007, opened 2026-09-03
fix(settings): recover from render failures

### LINKED ISSUE #1006: [Bug]: Settings render errors can blank the whole React root
### Describe the bug

On current `main`, the chat and resource surfaces are protected by `UiErrorBoundary`, but the Settings subtree is not. If any Settings child throws while rendering, the error escapes `WorkbenchSettingsStage` and reaches the React root, so the window can lose all rendered controls instead of showing a recoverable fallback.

This gap was found while investigating #1003. It does **not** establish that a React render error caused that one-off macOS 0.2.19 incident: #848 separately fixed the known Settings/WKWebView atomic-transition problem in v0.2.25, and the CLI recycle in #1003 is only temporally associated.

### To reproduce

1. Render `WorkbenchSettingsStage` with a `SettingsPage` child that throws once during render.
2. Observe that the exception escapes the Settings stage; there is no Settings-level error UI or retry action.

A focused Vitest/Testing Library regression can reproduce this deterministically without a native macOS build.

### Expected behavior

Settings render failures should stay contained within the Settings stage. The app should keep a visible, localized error panel with the error detail and a Retry action. Retrying after a transient fault should remount the Settings content. Navigating to another Settings route should also reset the boundary.

### Version

Current `main` (`0.2.30`).

### Operating system

Platform-independent React recovery gap; discovered from a macOS report.

YIKUAIBANZI:
## Summary

- Contain Settings child render failures in the existing `UiErrorBoundary` instead of letting them clear the whole React root.
- Keep a visible localized fallback with error detail and Retry across all 15 shipped locales.
- Reset the boundary when the Settings section/tab route changes.

## Root cause

`WorkbenchSettingsStage` rendered `SettingsPage` without an error boundary, while the chat and resource surfaces already had local boundaries. A Settings child render exception therefore escaped to the React root and left no recoverable controls.

This independently reproducible recovery gap was found while investigating #1003. It does **not** establish that the CLI upgrade caused that one-off macOS 0.2.19 incident; #848 separately fixed the known Settings/WKWebView atomic-transition issue in v0.2.25.

Fixes #1006
Related: #1003

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Validation

- `pnpm deps:check`
- `pnpm audit:prod`
- `pnpm typecheck`
- `pnpm test` — 566 files / 6927 tests passed
- `pnpm lint`
- `python3 scripts/check-code-quality-gates.py --mode final`
- `python3 scripts/publish-website-downloads.py --self-test`
- `pnpm build:ui`
- Focused regression was observed failing before the fix and passing afterward; it covers Retry and route-reset recovery.

Rust checks were not run locally because this host has no `cargo` or `rustc`. This PR changes no Rust files; GitHub CI remains the Rust verification path.

## Checklist

- [ ] I ran `cargo test` in `src-tauri` (or `cargo check` for docs-only) — unavailable on this host; no Rust changes
- [x] User-facing strings go through the locale catalogs
- [x] Changelog updated


RongleCat:
Thanks @YIKUAIBANZI — this is the right fix for #1006.

Same recovery landed via maintainer #1008 (squash on `main`) while this PR was open, so closing as a duplicate rather than re-applying.

Appreciate the focused Vitest coverage and the #1003 investigation notes.

PR #1010, opened 2026-09-03
feat(chat): typed tool presentation with explicit meta passthrough

shiaho777:
Introduce a pure tool presentation layer: recorded call fields derive a typed card (terminal/diff/read/search/web/generic) with bounded facts (query, result count/domains, path base). An explicit host meta field on tool events/segments/rows wins when present and is preserved across live updates, coalescing, weaving, and replay; otherwise the UI derives the same card so old journals keep a typed icon/label instead of bare generic. Activity rail prefers the typed facts with existing string parsing as fallback.

### DIFF
--- a/src/lib/grokActivitySteps.ts
+++ b/src/lib/grokActivitySteps.ts
@@ -21,6 +21,10 @@ import {
   toolPathBase,
   type ToolDisplayKind,
 } from "./toolDisplay";
+import {
+  resolveToolPresentation,
+  toolBucketForCard,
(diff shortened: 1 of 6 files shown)


RongleCat (approved):
## Review

Approved for squash-merge (maintainer batch — UI changes explicitly approved).

- Scope matches description; tests included; CI green.
- No secrets / window.confirm / AppWorkbench growth.
- Thanks @shiaho777

PR #1012, opened 2026-09-03
feat(chat): stable-tail incremental paint for streaming markdown

shiaho777:
Split long streaming markdown at a blank line outside fenced blocks so only the tail re-parses each tick; the stable prefix renders once as settled markdown. Falls back to single render for short sources, unclosed fences, or find-highlight mode, and settled turns always single-render so no mid-stream boundary persists.


RongleCat (approved):
## Review

Approved for squash-merge (maintainer batch — UI changes explicitly approved).

- Scope matches description; tests included; CI green.
- No secrets / window.confirm / AppWorkbench growth.
- Thanks @shiaho777

PR #1014, opened 2026-09-04
perf(chat): keep selection toolbar and composer poll off the hot path (#1013)

zhangxaochen:
Fixes #1013

Selecting chat text was committing `ConversationThread` on every `selectionchange` while dragging. Composer slash/@ rAF also forced layout on those frames via `innerText`. Empty `sessionChanges ?? []` allocated a new array and busted row memos.

## Summary
- Quote toolbar state lives in `TranscriptSelectionToolbarHost`; skip `setState` while the pointer is down; rAF + equality on commit
- Pause slash/@ DOM probe when the window is hidden or the selection is outside the composer (`useComposerController`)
- Reuse `EMPTY_SESSION_FILE_CHANGES` so empty change lists keep identity
- AppWorkbench lines decreased (rAF moved out of the freeze shell)

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore

## Test plan
- [x] `pnpm exec vitest run src/lib/transcriptSelectionBar.test.ts src/lib/composerLiveProbe.test.ts src/lib/atFileQuery.test.ts src/lib/sessionChanges.test.ts` (52 passed)
- [x] `pnpm exec eslint` on touched files
- [x] `python scripts/check-code-quality-gates.py --mode final` — `APP_NO_GROW` pass (15239 < 15350). Remaining fail is pre-existing `files_ge_1000` (79 > 77)
- [ ] `pnpm typecheck` currently red on `main` (`remark-math` / `rehype-katex` types) — not introduced here
- [ ] Manual: drag-select in a long assistant message — toolbar appears on pointer-up; Shift+arrows still show it; comment box keeps the bar when native selection collapses

## Checklist
- [x] Scoped tests + eslint on touched files
- [x] `cargo test` N/A (TS-only)
- [x] No new i18n copy
- [x] No `window.confirm` / `prompt` / `alert`
- [x] CHANGELOG Unreleased updated
- [x] No secrets included

PR #1016, opened 2026-09-04
fix(composer): quote Enter send + shared send-key pref (#1015)

### LINKED ISSUE #1015: [UX]: Quote comment Enter should match composer send, and Enter should send a quote-only draft
### Problem / motivation

After adding a transcript quote (with an optional comment), focus moves to the composer. Enter does not send if you type nothing else, even though the Send button already treats a quote card as a body.

Separately, the quote comment box hardcoded Ctrl+Enter to add the quote, ignoring **Settings → Chat → Send message with** (Enter vs Ctrl+Enter).

### Proposed solution

1. Keyboard submit uses the same "has body" rule as Send: draft, files, **or quote cards**.
2. The quote comment box follows `composerSendKey` (Enter vs Ctrl+Enter). A checkbox on the bar toggles that same preference.

### Additional context

Split out of #1014 / #1013 (those stay performance-only).

zhangxaochen:
Fixes #1015

Split from #1014 — quote submit UX, not the selection-toolbar perf work.

## Summary
- Enter in the composer sends a quote-only draft (Send button already did)
- Quote comment box uses the **same** Enter / Ctrl+Enter preference as the composer (Settings → General → Composer)
- The bar shows a hint for the current chord (no checkbox)

## Type of change
- [x] Bug fix

## Test plan
- [x] `pnpm exec vitest run src/lib/composerQuotes.test.ts src/lib/composerSendKey.test.ts src/i18n/messages.test.ts`
- [ ] Manual: add quote with comment, focus composer, Enter with empty input → sends
- [ ] Manual: change send-key in Settings; comment box Enter vs Ctrl+Enter matches the composer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,10 @@ See `docs/llm-wiki/release.md`.
 ## [Unreleased]
 
 ### Changed
-- The quote comment box uses the same Enter / Ctrl+Enter send shortcut as the composer. A checkbox on the bar toggles it.
+- The quote comment box submits with Enter (Shift+Enter for a new line). It does not follow the composer send-key setting.
 
 **中文 · 变更**
-- 划词评论框与输入框共用 Enter / Ctrl+Enter 发送快捷键。栏上的复选框可切换。
+- 划词评论框用 Enter 提交（Shift+Enter 换行）。不跟输入框的发送快捷键设置走。
 
 ### Fixed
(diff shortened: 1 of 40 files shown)


RongleCat:
CI `frontend` fails code-quality gates (`APP_NO_GROW` / AppWorkbench line ceiling). Functional quote Enter change looks good, but the AppWorkbench delta needs to stay at or below the freeze ceiling before merge.

RongleCat:
Superseded by #1028 (rebased onto main after #1014 Host extraction).

PR #1027, opened 2026-09-05
perf(chat): default-collapse tools to cut WebContent RAM (#1018)

RongleCat:
Partial fix for #1018

## Summary
- Tools no longer auto-expand while running (keeps activity VirtualList windowing)
- Streaming thoughts still open live via `openWhileRunning`
- Collapsed `TimelineToolRow` skips eager `toolExpandBody` string work

## Still open after merge
Journal/React still retain full `tool.output` in heap when collapsed — follow-up spill/hydrate needed. Keep #1018 open or spawn a follow-up for that slice.


RongleCat:
Superseded by #1028 (partial #1018 fix landed there).

PR #1045, opened 2026-09-07
fix(chat): keep ordinary JSON fences in the transcript

jay3cx:
## Why

Assistant replies that include a ` ```json ` sample (API payloads, configs, etc.) never showed the JSON in the bubble. The model wrote it; the journal stored it; the UI stripped it.

`extractAutomationPayload` uses ` ```json ` as a fallback for schedule-task fences, then **always** deleted every matching fence from `cleanText` — including JSON that is not a scheduled task (`title` + `prompt` + `frequency`).

## Fix

- Strip ` ```grok-automation ` always (internal protocol).
- Strip ` ```json ` only when it parses as a real automation config.
- Leave ordinary JSON samples in the transcript.

## Test

`pnpm exec vitest run src/lib/automationSetup.test.ts`

Adds coverage that a Databus-style JSON example stays visible, while schedule JSON / broken grok-automation fences still disappear.

## Workaround until this ships

Copy the assistant message, or ask the model not to wrap the sample in a json fence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Fixed
+- JSON examples in chat stay visible instead of being swallowed as schedule fences.
+
+**中文 · 修复**
+- 对话里的普通 JSON 示例会显示出来，不再被当成定时任务围栏剥掉。
+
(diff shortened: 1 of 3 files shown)


jay3cx:
Closing — will ship this as a local package instead of an upstream PR.

PR #1040, opened 2026-09-06
fix(linux): prefer host WebKitGTK in AppImage to avoid SIGBUS on quit

Facilitat-dev:
## Summary
Linux AppImage was SIGBUS-ing leftover `WebKitNetworkProcess` when the squashfs unmounted (`BUS_ADRERR` in bundled `libsqlite3`, same second as FUSE teardown). Related to the Hyprland/AMD `EGL_BAD_PARAMETER` black window (#539).

When host WebKitGTK 4.1 is present, re-exec against it so helpers are not mapped from the AppImage. Otherwise wait for AppImage-backed helpers before exit. Opt out: `GROK_SKIP_SYSTEM_WEBKIT=1`.

Verified on Arch + Hyprland: extract/system-WebKit launcher keeps `WebKitNetworkProcess` and `WebKitWebProcess` on `/usr/lib/webkit2gtk-4.1`, UI loads, no new coredump.

## Type of change
- [x] Bug fix

## Checklist
- [x] I ran `cargo test` in `src-tauri` (`linux_webkit` unit tests)
- [x] `cargo clippy --lib -- -D warnings`
- [x] Docs / `docs/BUILD.md` / CHANGELOG [Unreleased] updated
- [x] No secrets included

Issue #539, opened 2026-08-08
[Bug][Linux] AppImage black window on Hyprland/AMD — WebKit EGL_BAD_PARAMETER


FrancoEscob:
## Summary

On **Arch Linux + Hyprland (Wayland) + AMD GPU**, the official **Linux AppImage** opens a window that stays **fully black**. The Rust host process keeps running (media server, ACP/agent prewarm, auth can succeed), but the **WebKitGTK WebView never paints**.

This looks like a **packaging / bundled WebKit** issue on modern Wayland stacks, not a missing CLI or bad install.

## Environment

| Item | Value |
|------|--------|
| Grok App | **v0.2.11** AppImage (`Grok_0.2.11_amd64.AppImage`) |
| OS | Arch Linux |
| Compositor | Hyprland **0.56.1** (Wayland) |
| Session | `XDG_SESSION_TYPE=wayland`, XWayland enabled (`DISPLAY=:0`) |
| GPU | AMD Radeon Graphics (Cezanne / Renoir, `radeonsi`, Mesa **26.1.6**) |
| System WebKit | `webkit2gtk-4.1` **2.52.5-2** (used only in workaround) |
| Grok Build CLI | **1.0.0** (works; already signed in) |

## Steps to reproduce

1. Download AppImage from [Releases](https://github.com/RongleCat/grok-app/releases) v0.2.11.
2. `chmod +x Grok_0.2.11_amd64.AppImage && ./Grok_0.2.11_amd64.AppImage`
3. Observe a **black** window titled `Grok` (WM class `Grok-app`).

Also tried (still black with stock AppImage WebKit):

```bash
WEBKIT_DISABLE_DMABUF_RENDERER=1 ./Grok_0.2.11_amd64.AppImage
WEBKIT_DISABLE_DMABUF_RENDERER=1 WEBKIT_DISABLE_COMPOSITING_MODE=1 ./Grok_*.AppImage
WEBKIT_DISABLE_DMABUF_RENDERER=1 GDK_BACKEND=x11 ./Grok_*.AppImage
# plus pure X11 (unset WAYLAND_DISPLAY), software GL (LIBGL_ALWAYS_SOFTWARE=1), etc.
```

Note: the AppImage already forces `GDK_BACKEND=x11` via `apprun-hooks/linuxdeploy-plugin-gtk.sh` (Tauri Wayland issue reference).

## Actual behavior

- Window is ~100% near-black (`#0D0D0D`); no welcome/setup UI.
[...]

PR #1066, opened 2026-09-07
fix(linux): prefer host WebKitGTK in AppImage to avoid SIGBUS on quit

RongleCat:
## Summary
Maintainer rebase of community PR #1040 (@Facilitat-dev) onto current `main`.

When host WebKitGTK 4.1 is present, re-exec against it so AppImage helpers are not mapped from the squashfs; otherwise wait for AppImage-backed helpers before exit. Opt out: `GROK_SKIP_SYSTEM_WEBKIT=1`.

Fixes the AppImage FUSE teardown `SIGBUS` / `BUS_ADRERR` class related to #539.

## Attribution
Original work by @Facilitat-dev in https://github.com/RongleCat/grok-app/pull/1040

## Test plan
- [x] `pnpm exec vitest run src/lib/whatsNew.test.ts`
- [ ] CI Rust Linux
- [x] `linux_webkit` is `cfg(target_os = "linux")` only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,12 @@ See `docs/llm-wiki/release.md`.
 ### Fixed
 - Opening a file in Review keeps the macOS title bar visible. Focus scrolls only inside Review (#1041).
 - Advanced context-window Save stays fully clickable. The flyout no longer clips the button (#1047).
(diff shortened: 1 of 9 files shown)


RongleCat:
Merged — thank you @Facilitat-dev for the original #1040 fix.

PR #1131, opened 2026-09-09
fix(deps): pin katex to the version rehype-katex renders with

shiaho777:
## Summary
The direct `katex` dependency (`^0.18.4`) exists only to load `katex/dist/katex.min.css` ([`markdownMath.ts`](src/lib/markdownMath.ts#L19)); the actual math HTML is rendered by `rehype-katex`, which resolves **katex 0.16.47** (mermaid and remark-math's transitive copies agree). The app has been shipping a **CSS/JS version skew** — 0.18.4 styles with 0.16.47 rendering JS — held together only by KaTeX's stable class names; any class rename between those lines would silently break math styling.

This pins the direct dep to `^0.16.47`, so CSS and JS now come from a single version and the duplicate katex 0.18.4 install (~4.3 MB) disappears from the tree:

```
before: katex 0.18.4 (direct, CSS) + katex 0.16.47 (rehype-katex/remark-math/mermaid)
after:  katex 0.16.47 (single instance)
```

**Stacked on #1135 (fmt/clippy repair) so the Rust CI leg is green; once #1135 merges this diff is just the two real files.

## Type of change
- [x] Bug fix
- [ ] Documentation
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (637 files / 7411 tests green)
- [x] Docs-only Rust impact: `cargo check` n/a — no Rust change
- [x] User-facing strings go through `src/i18n/messages.ts` — CHANGELOG entry only (en + zh, first sentence ≤ 90 chars)
- [x] Docs / `docs/llm-wiki` updated if behavior changed — behavior unchanged (version alignment)
- [x] No secrets included

## Verification
- `pnpm why katex` → single `katex 0.16.47` across direct + rehype-katex + remark-math + mermaid
- Full local CI suite green: typecheck / vitest / lint / quality gates / website self-test / `build:ui`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,18 @@ See `docs/llm-wiki/release.md`.
 
 ## [Unreleased]
 
+### Fixed
+- Math formulas render with one matching KaTeX version again. The bundled CSS had drifted ahead of the JS.
+
+**中文 · 修复**
+- 数学公式恢复 CSS 与 JS 同版本渲染，不再出现样式超前于脚本。
+
 ### Changed
 - The theme editor modal loads on demand instead of joining app startup.
 
 **中文 · 变更**
 - 主题编辑器改为按需加载，不再拖累应用启动。
+
 
 ## [0.2.34] - 2026-09-09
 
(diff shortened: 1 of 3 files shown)


RongleCat:
Rebased onto latest `main` (post-#1132) to resolve Unreleased CHANGELOG conflict with the theme-editor lazy-load bullet. CI re-running.

PR #1135, opened 2026-09-09
style(session): repair rustfmt + clippy regressions on main (0.2.34 hunks)

shiaho777:
## Summary
The 0.2.34 freeze-remediation commits (`a5822341`, `ada505bc`) landed with rustfmt and clippy violations, so the **Rust CI leg fails on current main** and every open PR inherits the red:

- **rustfmt**: the retry-emit `live_sid` chain (`a5822341`) plus new hunks in `control.rs` / `events.rs` / `events_bg.rs` / `turn.rs` / `image_thumb.rs` (`ada505bc`)
- **dead_code**: `read_remote_image_body` is only used by tests — gate it `#[cfg(test)]`
- **map_identity**: two `.map_err(|e| e)` no-ops in `image_thumb.rs`
- **large_enum_variant**: `PendingSessionPersist` variants were 1184B / 568B — both payloads boxed, enum drops to 16B

No semantic change; green Rust CI is the point. Same class as #1130 (the `whatsNew` leg that release also fixed).

`cargo fmt --check` / `cargo clippy --all-targets -- -D warnings` / `cargo test` (1,846 tests) all green on this branch.

**This PR is the base of the stack** — merge it first and #1131 / #1132 / #1133 / #1134 / #1136 / #1137 / #1145 all collapse to their own single-commit diffs.

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Refactor / chore (CI hygiene)

## Checklist
- [x] I ran `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` in `src-tauri` (all green)
- [x] No frontend change (frontend untouched)
- [x] No user-facing strings
- [x] No secrets included

PR #1134, opened 2026-09-09
chore(deps): remove unused @vitejs/plugin-react devDependency

shiaho777:
## Summary
The build uses `@vitejs/plugin-react-swc`; the Babel-based `@vitejs/plugin-react` has zero references in the repo (only the `package.json` entry). This drops the dead devDependency (~90 lockfile lines) and fixes the stale comment in `vite.config.ts` that still pointed at "App.tsx ~800KB+" — the module that motivated SWC is now `AppWorkbench.tsx`.

**Stacked on #1135 → #1132 → #1133 so CI stays green; once those merge the diff is the two files.

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (638 files / 7417 tests green)
- [x] No Rust change — frontend suite + gates + `build:ui` green locally
- [x] User-facing strings — none (internal dependency cleanup)
- [x] No `window.confirm` / `prompt` / `alert`
- [x] Docs / `docs/llm-wiki` — no behavior change
- [x] No secrets included

PR #1133, opened 2026-09-09
perf(build): ship KaTeX fonts as woff2 only

shiaho777:
## Summary
KaTeX emits every `@font-face` with three `src` entries — woff2 + woff + ttf. Grok App's target WebViews (WKWebView / WebView2 / WebKitGTK) all load woff2, so the fallback formats were dead weight: **~700KB / 40 font asset files** in every dist build.

This adds a small `generateBundle` Vite plugin (`katex-woff2-only`) that:
- strips the non-woff2 `url(...)` src entries from bundled CSS (`stripKatexFallbackFontSrc`)
- deletes the matching `KaTeX_*.woff` / `KaTeX_*.ttf` assets from the bundle

The pure helpers live in [`src/lib/katexFontTrim.ts`](src/lib/katexFontTrim.ts) with unit tests (bundled-asset naming, quoted urls, multi-font-face sheets, non-KaTeX fonts left intact). Verified on `build:ui` output:

```
before: 19 woff2 + 20 woff + 20 ttf KaTeX font assets
after:  19 woff2 +  0 woff +  0 ttf
```

CSS contains zero `format("woff")` / `format("truetype")` references afterwards; every font-face keeps its woff2 entry, so rendering is unchanged.

**Stacked on #1135 → #1132 so CI stays green; once those merge the diff is just the four files.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / chore (build size, no behavior change)

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (638 files / 7417 tests green, including new `katexFontTrim.test.ts`)
- [x] No Rust change — frontend suite + gates + `build:ui` green locally
- [x] User-facing strings via `src/i18n/messages.ts` — CHANGELOG entry only (en + zh)
- [x] No `window.confirm` / `prompt` / `alert`
- [x] Docs / `docs/llm-wiki` — no behavior change (build output only)
- [x] No secrets included

PR #1132, opened 2026-09-09
perf(app): lazy-load the theme editor modal from the sidebar

shiaho777:
## Summary
`WorkbenchSidebar` statically imported `ThemeEditorModal`, pulling the appearance-editor module closure into the boot-critical App chunk even though the modal only mounts when the user opens it from the sidebar rail. The bundle audit found this was the last big lazy-loading win on the startup path.

This switches the import to `React.lazy` + `Suspense fallback={null}` (same pattern as the app's other one-shot modals). Verified in `build:ui` output:

- `ThemeEditorModal-*.js` and `useAppearanceEditorModel-*.js` now emit as separate on-demand chunks
- the App chunk only carries the dynamic-import reference (no modal code)
- `index.html` does not preload the modal

**Stacked on #1135 (fmt/clippy repair) so CI is green; once that merges this diff is just the two files.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / chore (perf, no behavior change)

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (637 files / 7411 tests green)
- [x] No Rust change — frontend suite + gates + `build:ui` green locally
- [x] User-facing strings via `src/i18n/messages.ts` — CHANGELOG entry only (en + zh)
- [x] No `window.confirm` / `prompt` / `alert`
- [x] Docs / `docs/llm-wiki` — no behavior change; startup strategy matches existing lazy-modal pattern
- [x] No secrets included

## Regression notes
First open of the theme editor now pays a local-disk chunk load (tens of ms in Tauri asset protocol) — same trade-off the settings stage, Office previews, and pet window already make.

PR #1136, opened 2026-09-09
chore(deps): drop unused thiserror; declare only used tokio features

shiaho777:
## Summary
Two manifest-hygiene fixes found in a dependency audit:

1. **`thiserror = "2"` has zero usages** — declared since the first commit, never imported or derived anywhere (the crate's error strategy is `Result<T, String>` by design). Dead dependency removed.

2. **`tokio` moves off `"full"`** onto the exact feature set the crate uses: `macros`, `rt-multi-thread`, `sync`, `time`, `process`, `io-util`, `net`. Verified by enumerating every `tokio::…` path in `src/`:
   - `tokio::fs` — zero call sites (file IO is `spawn_blocking` + `std::fs` throughout, 225+ sites)
   - `tokio::signal` — zero call sites
   
   Cargo feature unification with tauri/axum still applies; this makes the manifest stop promising modules the crate never calls. A comment marks the rule for future call sites.

**Kept on purpose:** `reqwest`'s `blocking` feature — actively used by `mcp_oauth`, `session_api`, and `proxy` (the blocking client is the correct tool in their sync/spawn_blocking contexts; converting them is a behavioral refactor outside this PR's scope).

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Refactor / chore

## Checklist
- [x] I ran `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` in `src-tauri` (all green on this branch)
- [x] No frontend change — frontend tree untouched
- [x] No user-facing strings
- [x] Docs / `docs/llm-wiki` — no behavior change
- [x] No secrets included

**Stacked on #1135 → #1132 → #1133 → #1134 so CI stays green (main currently fails fmt+clippy; #1135 fixes that).


RongleCat:
Rebased onto current `main` (post-#1138) by maintainer: dropped the superseded rustfmt commit (#1135 / already in #1138), and moved CHANGELOG bullets into `## [Unreleased]` so we do not rewrite shipped `0.2.34`.

CI should re-run on the force-push tip. Will squash-merge bottom-up once green.

PR #1137, opened 2026-09-09
perf(host): run settings store IO on the blocking pool in commands

shiaho777:
## Summary
Tauri commands called `store::load_settings()` / `store::save_settings()` directly on their async worker. The settings file sits behind an **exclusive store lock**, and `load_settings` may even rewrite it during one-time migrations — so a contended lock blocks the async worker, the same freeze class the Windows remediation wave removed from session-map locks (cf. `git_status` moving to the blocking pool in #990).

This adds `store::load_settings_async` / `store::save_settings_async` (thin `spawn_blocking` delegations) and converts the **command entry points** in `src/commands/`:

- **42 call sites** now go through the async accessors (verified by the compiler)
- **12 sites** correctly stay sync — they already run inside `spawn_blocking` closures and were initially misconverted; `cargo check` caught every one and they were reverted
- **7 sites** in sync `*_blocking` helpers (already pool-only) stay sync
- one closure call site (`cli_install_latest`) was restructured to a `match` that keeps the lazy-load semantics

Out of scope (deliberately): the ~118 deeper internal call sites in `session_manager`, `serve.rs`, background workers — those run in mixed contexts that need case-by-case analysis, not a drive-by.

## Type of change
- [x] Bug fix (reliability — async workers no longer block on the settings file lock)
- [ ] New feature
- [ ] Refactor / chore

## Checklist
- [x] I ran `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` in `src-tauri` (all green; 1,841 tests)
- [x] I ran `pnpm typecheck` / `pnpm test` (638 files) / `lint` / gates / `build:ui` (CHANGELOG entry only on the frontend side)
- [x] User-facing strings — CHANGELOG entry only (en + zh)
- [x] No secrets included

**Stacked on #1135 → #1132 → #1133 → #1134 → #1136 so CI stays green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,12 @@ See `docs/llm-wiki/release.md`.
 ### Changed
 - The theme editor modal loads on demand instead of joining app startup.
 - Bundled KaTeX math fonts ship as woff2 only, trimming the install size.
+- Settings reads and writes run on the blocking pool, off the async command path.
 
 **中文 · 变更**
 - 主题编辑器改为按需加载，不再拖累应用启动。
 - 打包内的 KaTeX 数学字体只保留 woff2 格式，安装包更小。
+- 设置的读写改走阻塞线程池，异步命令不再被设置文件锁卡住。
 
 
 ## [0.2.34] - 2026-09-09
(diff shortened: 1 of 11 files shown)


RongleCat:
Rebased onto current `main` (post-#1138) by maintainer: dropped the superseded rustfmt commit (#1135 / already in #1138), and moved CHANGELOG bullets into `## [Unreleased]` so we do not rewrite shipped `0.2.34`.

CI should re-run on the force-push tip. Will squash-merge bottom-up once green.

PR #1145, opened 2026-09-10
fix(whatsNew): parse the full CHANGELOG section, not up to the first Z

shiaho777:
## Summary
`extractVersionBody` ended its section lookahead with `\\Z`, intended as "absolute end of input" (PCRE/Python semantics). **In JS regex, `\\Z` is a literal-Z identity escape.** So any version section containing a capital Z parsed only up to that character, and the What's New popup silently dropped every entry after it.

Real impact on current CHANGELOG:
- **Unreleased** truncates at the "ZDR…" line in Fixed — everything after it (rest of Fixed + entire Added / Changed sections, English and Chinese) never reached the popup or the length guard
- Shipped sections 0.2.27 ("Zhipu GLM"), 0.2.24 ("*Zug*"), 0.2.21 ("ZWSP"), 0.2.13 ("A–Z"), 0.2.2 ("Zen mode") were truncated the same way while they were current

This replaces `\\Z` with `(?![\\s\\S])` (the JS way to assert absolute end-of-input) and adds a regression test with Z-bearing entries mid-section. Verified the full Unreleased now parses clean against the popup's ≤90-char first-sentence rule.

Found while verifying the CHANGELOG length guard during the #1131–#1137 batch — a 96-char line passed the test only because it sat after a "ZDR" line the parser never reached.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore

## Checklist
- [x] I ran `pnpm typecheck` and `pnpm test` (whatsNew suite green incl. the new regression test)
- [x] No Rust change — frontend only
- [x] User-facing strings via CHANGELOG only (en + zh, first sentence ≤ 90 chars)
- [x] No secrets included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,13 @@ See `docs/llm-wiki/release.md`.
 - Math formulas render with one matching KaTeX version again. The bundled CSS had drifted ahead of the JS.
 - Wallpaper no longer flashes black while streaming or following the chat tail.
 - Imagine portrait thumbnails stay contained when the wallpaper window narrows.
+- What's New now lists every entry; some were cut off before.
 
 **中文 · 修复**
 - 数学公式恢复 CSS 与 JS 同版本渲染，不再出现样式超前于脚本。
 - 流式输出和自动跟随聊天底部时，静态及动态壁纸不再闪黑。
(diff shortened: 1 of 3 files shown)


RongleCat:
Rebased onto current `main` (post-#1144) by maintainer; CHANGELOG conflict resolved. CI re-running.

PR #1156, opened 2026-09-10
refactor(settings): declare the model bag honestly; drop the Record casts

shiaho777:
## Summary
`SettingsViewModel` carried a `[key: string]: unknown` catch-all, and nine section components widened it further with `as SettingsViewModel & Record<string, any>` — a double escape hatch that made the entire settings tree compiler-invisible (any field typo anywhere in settings rendered as `undefined` at runtime).

Two coupled changes (they are one seam — declaring without de-casting, or de-casting without declaring, each leaves the tree red):

1. **Declare the 107 fields** the sections actually destructure, on `SettingsViewModel` (the context view — they belong to the page's internal state, not `SettingsPageProps`). Types were **inferred from the real SettingsPage construction-site sources via the TS checker** (`useState` locals, memoized presenters, setters), so every declared type matches what the page actually provides.
2. **Delete the cast** from all nine sections (`AboutSection`, `AccountSection`, `ArchivedSection`, `GeneralSection`, `RemoteImSection`, `ShortcutsSection`, `RuntimeSection`, `AppearanceSection`, `ExtensionsSection`).

Audit bycatch: two suspected fields (`lightFrom`/`darkFrom`) turned out to be locals destructured from `themeSchedule` — not model fields, correctly left out.

Ratchet: `SETTINGS_ANY_CAST_CEILING` 9 → **0** — settings sections can no longer widen the model bag past what `types.ts` declares. (Third ratchet extinction: `LOOSE_PROPS` hit 0 in #1155; only `TS_NOCHECK` remains at 1.)

**Stacked on #1150 → #1151 → #1152 → #1153 → #1155** — the field declarations are required fields, which is only sound on top of the typed SettingsStage props (the any-bag stage could not satisfy them).

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore

## Checklist
- [x] `pnpm typecheck` / `pnpm test` (641 files) / `lint` / gates / self-test / `build:ui` — all green
- [x] No Rust change
- [x] No user-facing strings / no visual change
- [x] No secrets included

PR #1155, opened 2026-09-10
refactor(app): type WorkbenchSettingsStage props (223 fields) — last loose bag

shiaho777:
## Summary
The **last loose props bag in the workbench**: the settings stage received 223 props typed `{ [key: string]: any }` — 79 setters, 32 callbacks, 111 values, all invisible to `tsc`.

This replaces it with an explicit 223-field interface, **inferred from the real call-site types via the TS checker**, living in its own module [`src/app/workbenchSettingsStageProps.ts`](src/app/workbenchSettingsStageProps.ts) — the first draft put it inline and tripped the `files_ge_1000` gate (the ratchet working as designed), so the contract now has its own importable home (usable by tests too).

Details:
- Oversized inline shapes use named types: `CliInfoState`, `CompactionModeId` / `CompactionDetailId`, a `TrayHandlers` alias
- One wrong inline annotation (`setSetup` callback) fell away with real types in scope

Ratchet: `LOOSE_PROPS_CEILING` 1 → **0 — loose props bags are extinct in `src/`**. Any new `[key: string]: any` Props now fails the quality gate, closing the door behind the cleanup.

**Stacked on #1150 → #1151 → #1152 → #1153.**

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore

## Checklist
- [x] `pnpm typecheck` / `pnpm test` / `lint` / gates (`files_ge_1000` back to 80, `loose_props` 0) / self-test / `build:ui` — all green
- [x] No Rust change
- [x] No user-facing strings / no visual change
- [x] No secrets included

PR #1153, opened 2026-09-10
refactor(app): type the composer column/shell props seam (192 fields)

shiaho777:
## Summary
`WorkbenchComposerColumn` received a **192-prop any bag** and re-spread the whole thing into `WorkbenchComposerShell` — the largest prop seam in the workbench was entirely compiler-invisible.

This gives ComposerColumn an explicit 192-field Props interface, **inferred from the real call-site types via the TS checker**, and aliases the Shell's Props to it: the shell consumes the same bag via `{...p}` spread, so **one interface now guards both ends** of the seam.

Details:
- Oversized inline shapes collapse into local aliases (`ComposerAttachLabels`, `PromptHistoryEntryOpts`, `SendQueueApi = ReturnType<typeof useSendQueue>`)
- 22 redundant inline `(x: any)` annotations in the shell fall away — the file is now `any`-free
- Typing surfaced that `welcomeProviderBrandNode` is a `ReactNode` (not `JSX.Element`) at the source — now encoded correctly
- Five props the checker could not infer at the call site (`openQueueEdit`, `queueEditItemId`, `requestClearSendQueue`, `setCompactNote`, `setShowCompactModal`) were resolved against their hook sources (`useQueueEditDialog`, `useCompactDialog`)

Ratchet: `LOOSE_PROPS_CEILING` 3 → 1 — only `WorkbenchSettingsStage` remains loose (next PR).

**Stacked on #1150 → #1151 → #1152.**

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore

## Checklist
- [x] `pnpm typecheck` / `pnpm test` (641 files) / `lint` / gates / self-test / `build:ui` — all green
- [x] No Rust change
- [x] No user-facing strings / no visual change
- [x] No secrets included

PR #1152, opened 2026-09-10
refactor(app): type WorkbenchChatStage props (91 fields)

shiaho777:
## Summary
Replaces `WorkbenchChatStage`'s `{ [key: string]: any }` bag with an explicit 91-field Props interface, **inferred from the real call-site types via the TS checker**. This is the chat-surface seam (ConversationThreadLive, plan/status bars, chat find, tasks panel, error banner) — prop typos now fail `tsc` instead of surfacing at runtime as undefined chrome.

Typing surfaced two real shapes worth encoding (both verified against their sources):
- `structuredOutputUsage` is nullable at the source (`… | null`)
- `markSessionWorktree` takes `string | null | undefined` ids per the hook's actual signature

Plus a local `StreamStallView` alias documenting the nullable stall state, and `resolveLocale()` at the `ConversationThreadLive` boundary (it wants `Locale`, the prop is a plain string).

Ratchet: `LOOSE_PROPS_CEILING` 4 → 3.

**Stacked on #1150 → #1151** (gates + floating menus) — the chain keeps each PR one concern.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor / chore

## Checklist
- [x] `pnpm typecheck` / `pnpm test` (641 files) / `lint` / gates / self-test / `build:ui` — all green
- [x] No Rust change
- [x] No user-facing strings / no visual change
- [x] No secrets included

PR #1157, opened 2026-09-10
refactor(session): type the host-events ctx; drop @ts-nocheck

shiaho777:
## Summary
The entire backend-event reduction layer — **33 Tauri channels** (`session://stream`, `state`, `tool`, `permission`, `ask_user`, `plan`, …), the streaming projection / permission gates / plan merging machinery — ran under `// @ts-nocheck`, with every AppWorkbench binding flowing from a single `ctxRef.current as any` cast. This was the last typing escape hatch in `src/`.

This gives `SessionHostEventsCtx` an explicit 53-field interface, **inferred from the real call-site types via the TS checker**, and deletes the directive and the cast.

**Typing the seam surfaced three latent bugs — all fixed in this PR:**
1. `GoalOrchHostPayload` was **used but never imported** (line 1414) — masked by the nocheck for an unknown time
2. `PlanState` was **used but undefined anywhere in the tree** — silently `any`; it is `SessionPlanState`, now spelled correctly
3. Two dead props were threaded into the hook and never read (`setTurnStartedAt`, `maxConcurrentAgents`) — dropped from the interface and the call site

Ratchet: `TS_NOCHECK_CEILING` 1 → **0**. With this, **all three typing ratchets sit at zero** (loose props → #1155, settings any-casts → #1156, nocheck → here): the AppWorkbench decomposition's type-safety debt is repaid, and the gates make regression structurally impossible.

**Stacked on #1150 → #1151 → #1152 → #1153 → #1155 → #1156.**

## Type of change
- [x] Bug fix (two latent bugs + dead code removal)
- [ ] New feature
- [x] Refactor / chore

## Checklist
- [x] `pnpm typecheck` / `pnpm test` (641 files) / `lint` / gates (all three typing ratchets = 0) / self-test / `build:ui` — all green
- [x] No Rust change
- [x] No user-facing strings / no visual change
- [x] No secrets included


RongleCat:
Fixed post-#1128 typecheck: added `gitBranches*` / `switchToBranch` / `refreshGitBranches` to `WorkbenchComposerColumnProps` so the typed seam matches the branch-chip feature.

PR #1173, opened 2026-09-11
feat(settings): add appearance toggle for chat virtual scrolling

KarnaughK:
## Summary

In some chats the virtual list’s height math is off, so scrolling up from the bottom jumps instead of moving 1:1 with the gesture.

Fixing that pin/browse window properly is a large change. This PR does not rewrite the virtualizer. It adds an opt-out in **Settings → Appearance → Interface → Chat scroll optimization**:

- **On by default** — same as today’s behavior (virtual window still turns on at the existing row/height thresholds).
- **Off** — persist locally (`localStorage`) and use native overflow for the conversation list.

That is enough for people who hit the jump and do not need virtualization (typical chats of ~20–30 turns, or machines that are fine mounting the full transcript). Native scroll also tends to feel better when the list is not huge.

## Type of change
- [x] New feature

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run src/lib/chatVirtualScrollPref.test.ts src/lib/settingsCatalog.test.ts src/hooks/useChatMessageVirtualizer.test.tsx src/i18n/messages.test.ts`
- [ ] Manual: Settings → Appearance → Interface — toggle is checked by default
- [ ] Manual: uncheck it, reload — preference sticks; scrolling up from the bottom no longer jumps
- [ ] Manual: check it again — long chats still virtualize as before


RongleCat:
Maintainer approved merge as an Appearance opt-out for virtual-scroll jump cases. Will squash-merge once CI is green on current main.

PR #1181, opened 2026-09-12
chore(ui): drop dead conversation wrapper and use-stick-to-bottom dep

shiaho777:
## Summary
- `src/components/ui/conversation.tsx` has zero importers — dead file
- `use-stick-to-bottom` was only imported by that file; the live chat thread uses the in-tree `src/hooks/useStickToBottom.ts`
- Removes both the file and the dependency

#### Test plan
- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `grep` confirms no remaining references to the file or package

Generated with [Devin](https://devin.ai)


RongleCat:
Merged — thanks @shiaho777!

PR #1192, opened 2026-09-12
chore(rust): narrow file-level #![allow(dead_code)] to item-level

shiaho777:
## Summary
- 45 files suppressed `dead_code` **file-wide** with `#![allow(dead_code)] // residual-clippy` comments — which also silently hid any *new* dead code those files accumulated afterward.
- This PR removes every file-level suppression and annotates the actual dead items instead: ~150 targeted `#[allow(dead_code)]` on the specific fns / struct fields / methods rustc flags (protocol-parity DTOs, helpers retained for tests, platform-gated paths). Two files needed no annotation — their suppression was fully stale.
- Net effect: every file is now lint-guarded again; only the flagged items are exempt. Attribute placement only — zero behavior change.
- Pre-existing item-level and `cfg_attr` allows (`not(test)`, `target_os`) were left untouched.

## Test plan
- [x] `cargo check` — zero dead_code warnings (was hidden by blanket allows)
- [x] `cargo fmt --check` clean
- [x] `cargo clippy` — 0 warnings
- [x] `cargo test --lib` — 1866 tests pass

Generated with [Devin](https://devin.ai)

### DIFF
--- a/src-tauri/src/account.rs
+++ b/src-tauri/src/account.rs
@@ -4,7 +4,6 @@
 //! Billing is best-effort HTTP (same field shape as CLI `/usage` / billing extension).
 //! Heatmap + call logs are derived from local CLI session signals (and optional app journal).
 
-#![allow(dead_code)] // residual-clippy: billing helpers not yet wired in UI path
 use std::collections::{BTreeMap, HashSet};
 use std::fs;
 use std::io::{BufRead, BufReader};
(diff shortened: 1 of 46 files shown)


RongleCat:
CI red (clippy `-D warnings`): `badge_label_value` / `badge_count_value` never used after narrowing file-level `dead_code` allows. Please `#[allow(dead_code)]` on those items (or delete if truly dead) and re-push.

shiaho777:
Fixed: added `#[allow(dead_code)]` on `badge_label_value` / `badge_count_value` (they're only called under `#[cfg(target_os = "macos")]` / unix badge paths, so they warn on Linux CI). Also rebased onto current main and resolved the `acp_client.rs` conflict with the merged spawn_flags extraction — dead helpers that moved to `spawn_flags.rs` now carry the item-level allows there. `cargo clippy -- -D warnings` and `cargo fmt --check` are clean locally.

RongleCat:
Merged — thanks @shiaho777!

